PR #9, opened 2024-09-23
Prevent the \ character causing the version to wrap the next line

FinlayRJW:
## Before this PR
Previously if you finished a line in version.props with `\` it would wrap the version to include the following line

## After this PR
Updated the grammar such that version versions do not wrap to the following line so the `\` can be safely used

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
@@ -66,7 +66,7 @@ GROUP_PART=[^.:=\ \n\t\f\\] | "\\ "
 
 <WAITING_VALUE> {WHITE_SPACE}+                              { yybegin(WAITING_VALUE); return TokenType.WHITE_SPACE; }
 
-<WAITING_VALUE> {VALUE_CHARACTER}*                          { yybegin(YYINITIAL); return VersionPropsTypes.VERSION; }
+<WAITING_VALUE> {VALUE_CHARACTER}+                          { yybegin(YYINITIAL); return VersionPropsTypes.VERSION; }
 
 ({CRLF}|{WHITE_SPACE})+                                     { yybegin(YYINITIAL); return TokenType.WHITE_SPACE; }
 
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
@@ -33,8 +33,7 @@ import com.intellij.psi.TokenType;
 
 CRLF=\R
 WHITE_SPACE=[\ \n\t\f]
-FIRST_VALUE_CHARACTER=[^ \n\f\\] | "\\"{CRLF} | "\\".
-VALUE_CHARACTER=[^\n\f\\] | "\\"{CRLF} | "\\".
+VALUE_CHARACTER=[^ \n\f\\] | "\\"{CRLF} | "\\".
 END_OF_LINE_COMMENT=("#"|"!")[^\r\n]*
 COLON=[:]
 EQUALS=[=]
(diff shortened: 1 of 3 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
```diff
@@ -67,7 +66,7 @@ GROUP_PART=[^.:=\ \n\t\f\\] | "\\ "
 
 <WAITING_VALUE> {WHITE_SPACE}+                              { yybegin(WAITING_VALUE); return TokenType.WHITE_SPACE; }
 
-<WAITING_VALUE> {FIRST_VALUE_CHARACTER}{VALUE_CHARACTER}*   { yybegin(YYINITIAL); return VersionPropsTypes.VERSION; }
+<WAITING_VALUE> {VALUE_CHARACTER}+                          { yybegin(YYINITIAL); return VersionPropsTypes.VERSION; }
```
crogoz: I think it would be nice to add some tests for the lexer & parser. I see that Intellij has some frameworks to do that: https://plugins.jetbrains.com/docs/intellij/parsing-test.html and https://github.com/JetBrains/intellij-community/blob/idea/242.22855.74/platform/testFramework/src/com/intellij/testFramework/LexerTestCase.java.

I tested now the case where you have something like: 
```
com.fasterxml.jackson.core:jackson-databind = 2.17.2\
com.google.auto.service:auto-service = 1.1.1
```

I am seeing this syntax error (but it is quite hidden): `VersionPropsTokenType.COMMENT, VersionPropsTokenType.CRLF or VersionPropsTokenType.GROUP_PART expected, got '=' `.

Screenshot 2024-09-25 at 3 01 00 PM

With the new change from:https://github.com/palantir/gradle-consistent-versions-idea-plugin/pull/10/files#diff-b8f5308d750c8b00dcf2749190e1362cc968661cc7597f358224253cee97e9d5R69 we will silently fail - we won't run writeVersionsLock and the user won't see clearly that the versions.props file is invalid.

FinlayRJW: I'll look at adding some tests now

I originally added the check for bad character in the `wVL` change as previously `wVL` would run on every autosave which often coincided with half typed lines and cause lots of annoying errors, I wonder if the best thing would be a popup of some sort?

FinlayRJW: So I managed to fix the lexer so that it didnt include end of line comments and wrapping with \ but failed on the tests.

FinlayRJW:
I couldn't figure out how to get the https://plugins.jetbrains.com/docs/intellij/parsing-test.html to work with JUNIT5 I have still added a more manual test where we check the structure of the generated PSI tree which I think should do? @crogoz is it worth me continuing to dig into the intellij parsing-test or is this good enough?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
```diff
@@ -33,20 +33,18 @@ import com.intellij.psi.TokenType;
...
-VALUE_CHARACTER=[^\n\f\\] | "\\"{CRLF} | "\\".
-END_OF_LINE_COMMENT=("#"|"!")[^\r\n]*
+VALUE_CHARACTER=[^ \n\f]+
 COLON=[:]
 EQUALS=[=]
 DOT=[.]
-KEY_CHARACTER=[^:=\ \n\t\f\\] | "\\ "
-GROUP_PART=[^.:=\ \n\t\f\\] | "\\ "
+KEY_CHARACTER=[^:=\ \n\t\f]+
+GROUP_PART=[^.:=\ \n\t\f]+
```
crogoz, line 41: I think it doesn't parse correctly if you have spaces between the group parts eg. `com.palantir .conjure: conjure = 4.49.0` either and this might be due to line 72

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
```diff
@@ -33,20 +33,18 @@ import com.intellij.psi.TokenType;
 
 CRLF=\R
 WHITE_SPACE=[\ \n\t\f]
-FIRST_VALUE_CHARACTER=[^ \n\f\\] | "\\"{CRLF} | "\\".
-VALUE_CHARACTER=[^\n\f\\] | "\\"{CRLF} | "\\".
-END_OF_LINE_COMMENT=("#"|"!")[^\r\n]*
+VALUE_CHARACTER=[^ \n\f]+
```
crogoz, line 36: maybe rename this from VALUE_CHARACTER to VALUE/VERSION (since now VALUE_CHARACTER matches multiple characters) and you can update the lexer rule on line 68 to `<WAITING_VALUE> {VALUE_CHARACTER}`

crogoz: same for GROUP_PART lexter rule, I think it can become `<YYINITIAL> {GROUP_PART}` (line 49)

crogoz (approved):

crogoz:
re: parsing tests. 
I think it is fine for now then, we could create a github issue

PR #10, opened 2024-09-23
Smarter versions lock task running

FinlayRJW:
## Before this PR
Previously the file listener had a couple of annoying behaviours:
1. Would try and run on a malformed file (file which contains error elements) 
2. Would run even if the diff was the same (i.e. I delete a line then retype the exact same thing)
3. Always ran wVL as a separate task then refreshed the project

## After this PR
All of the above have been solved:
- Malformed files are detected and removed
- Now we are only looking at `VFileContentChangeEvent` so only a diff causes a run
- wVL is ran as part of the refresh for projects without `buildSrc`

## Possible downsides?
`buildSrc` seems to cause wVL to be ran like `./gradlew buildSrc:wVL` when ran from a project refresh. To prevent this we separately handle cases with `buildSrc`. We first run the task then refresh the project - this adds quite a bit of complexity and ideally we would run the task as part of refresh here as well

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
@@ -121,7 +121,10 @@ public final class VersionPropsFileListener implements AsyncFileListener {
 
     private void refreshProjectWithTask(Project project) {
         log.debug("Refreshing project {} with task {}", project.getName(), VersionPropsFileListener.TASK_NAME);
-        refreshProject(project, new ImportSpecBuilder(project, GradleConstants.SYSTEM_ID).withArguments(VersionPropsFileListener.TASK_NAME));
+        refreshProject(
+                project,
+                new ImportSpecBuilder(project, GradleConstants.SYSTEM_ID)
+                        .withArguments(VersionPropsFileListener.TASK_NAME));
     }
 
     private void refreshProject(Project project) {
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
@@ -45,6 +45,7 @@ import org.slf4j.LoggerFactory;
 
 public final class VersionPropsFileListener implements AsyncFileListener {
     private static final Logger log = LoggerFactory.getLogger(VersionPropsFileListener.class);
+    private static final String TASK_NAME = "writeVersionsLock";
 
     @Nullable
     @Override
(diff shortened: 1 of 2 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
```diff
@@ -62,24 +64,81 @@ public ChangeApplier prepareChange(List<? extends VFileEvent> events) {
...
                 .filter(project -> versionPropsEvents.stream()
-                        .anyMatch(event -> event.getPath().startsWith(project.getBasePath())))
+                        .anyMatch(event -> event.getPath().startsWith(project.getBasePath())
+                                && !isFileMalformed(project, event.getFile())))
                 .toList();
 
         return new ChangeApplier() {
             @Override
             public void afterVfsChange() {
+                String taskName = "writeVersionsLock";
```
crogoz: nit: I think you can move this to a static field

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
```diff
@@ -62,24 +64,81 @@ public ChangeApplier prepareChange(List<? extends VFileEvent> events) {
...
+        TaskCallback callback = new TaskCallback() {
+            @Override
+            public void onSuccess() {
+                log.debug("Task {} successfully executed", taskName);
+                refreshProject(project);
+            }
+
+            @Override
+            public void onFailure() {
+                log.error("Task {} failed", taskName);
```
crogoz: How does this look in the UI if it fails ? would it make sense to show a notification pop-up on the right side if the task fails (not necessarily for this PR though)

FinlayRJW: Currently if this fails the user sees the output of `wVL` which I think should be good enough for now - can think about pulling into a pop-up in the future will add to the quip

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
FinlayRJW: awesome have made it static - thanks for the review!

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
crogoz: ah nice, just checked how it looks! that is perfect then

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
```diff
@@ -62,24 +65,83 @@ public ChangeApplier prepareChange(List<? extends VFileEvent> events) {
...
+    private boolean hasBuildSrc(Project project) {
+        return Files.exists(Paths.get(project.getBasePath(), "buildSrc"));
+    }
+
+    private void runTaskThenRefresh(Project project) {
+        log.debug("Running task {} on project {}", VersionPropsFileListener.TASK_NAME, project.getName());
+        TaskCallback callback = new TaskCallback() {
+            @Override
+            public void onSuccess() {
+                log.debug("Task {} successfully executed", VersionPropsFileListener.TASK_NAME);
```
crogoz: TASK_NAME is sufficient, you don't need to prefix it with the class (same in the other places)
```suggestion
                log.debug("Task {} successfully executed", TASK_NAME);
```

crogoz (dismissed):

crogoz (dismissed):

crogoz (approved):

PR #11, opened 2024-09-23
Mark and pin the latest version during autocomplete suggestion

FinlayRJW:
## Before this PR
It has been very hard to know what the latest version of a particular package is when using auto complete

## After this PR
Now the latest version is pinned and labeled, you can also find the latest version by starting to type `latest`

I have also made it such that the autocompletes popup as soon as = has been pressed so you don't need to start typing the version 

## Possible downsides?
I could not find a way to **always** pin the latest version so say the latest is 2.0.0 and you start typing 1. it will not suggest 2.0.0

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
@@ -37,7 +37,6 @@ import org.slf4j.LoggerFactory;
 public class VersionCompletionContributor extends CompletionContributor {
 
     private static final Logger log = LoggerFactory.getLogger(VersionCompletionContributor.class);
-    private char previousChar = '\0';
 
     VersionCompletionContributor() {
         extend(
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
@@ -83,8 +83,6 @@ public class VersionCompletionContributor extends CompletionContributor {
 
     @Override
     public final boolean invokeAutoPopup(PsiElement position, char typeChar) {
-        boolean shouldInvoke = (previousChar == '=' && typeChar == ' ');
-        previousChar = typeChar;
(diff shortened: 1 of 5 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -62,7 +64,13 @@ public void addCompletions(
                                 .map(RepositoryExplorer::new)
                                 .flatMap(repositoryExplorer ->
                                         repositoryExplorer.getVersions(group, dependencyPackage).stream())
-                                .map(LookupElementBuilder::create)
+                                .map(version -> version.isLatest()
+                                        ? PrioritizedLookupElement.withPriority(
+                                                LookupElementBuilder.create(version)
+                                                        .withTypeText("Latest", true)
+                                                        .withLookupString("latest"),
+                                                Double.MAX_VALUE)
```
crogoz, line 72: one tiny nit: is it possible to keep the space between the `=` and the version ? 
at the moment I get `net.jodah:failsafe =2.4.4` instead of `net.jodah:failsafe = 2.4.4`

FinlayRJW: fixed

FinlayRJW: I also have a formatter in the works that will format on save

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -35,6 +37,7 @@
 public class VersionCompletionContributor extends CompletionContributor {
 
     private static final Logger log = LoggerFactory.getLogger(VersionCompletionContributor.class);
+    private char previousChar = '\0';
```
crogoz: let's remove this as well since it is not used anymore

FinlayRJW: good spot - sorry I missed that!

crogoz (approved):

PR #13, opened 2024-09-23
add caffeine cache to folders

FinlayRJW:
## Before this PR
Each call to get the folders required a call to the remote repo, this could be very expensive

## After this PR
Now a caffeine cache is used to store the getFolders results which can create a large speedup

## Possible downsides?
If a new package is added it wont appear until the cache expires - this was deemed acceptable as packages are added infrequently - although this is why we dont cache versions as they can change often


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -33,12 +36,24 @@ public class RepositoryExplorer {
     private static final Logger log = LoggerFactory.getLogger(RepositoryExplorer.class);
 
     private final String baseUrl;
+    private static final Cache<DependencyGroup, List<Folder>> folderCache;
```
crogoz: nit: I usually initialize the static fields directly in the declaration, makes it easy to read, esp. because the initialization logic is simple

FinlayRJW: good to know thanks!

crogoz (approved):

PR #17, opened 2024-09-26
lexer working

FinlayRJW:
## Before this PR
Had merged early before all issues resolved in https://github.com/palantir/gradle-consistent-versions-idea-plugin/pull/9

## After this PR
Now spaces around : and . cause syntax errors also generally tidied the lexer up


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
```diff
@@ -23,52 +23,52 @@ import com.intellij.psi.TokenType;
...
-<WAITING_VERSION> {EQUALS}                                  { yybegin(WAITING_VALUE); return VersionPropsTypes.EQUALS; }
+<YYINITIAL> {COMMENT}                            { yybegin(YYINITIAL); return VersionPropsTypes.COMMENT; }
+<YYINITIAL> {GROUP_PART}                         { yybegin(YYINITIAL); return VersionPropsTypes.GROUP_PART; }
+<YYINITIAL> {DOT}                                { yybegin(YYINITIAL); return VersionPropsTypes.DOT; }
+<YYINITIAL> {WHITE_SPACE}*{DOT}{WHITE_SPACE}*    { yybegin(INVALID_VALUE); return TokenType.BAD_CHARACTER; }
+<YYINITIAL> {COLON}                              { yybegin(WAITING_NAME); return VersionPropsTypes.COLON; }
+<YYINITIAL> {WHITE_SPACE}*{COLON}{WHITE_SPACE}*  { yybegin(INVALID_VALUE); return TokenType.BAD_CHARACTER; }
 
-<WAITING_VERSION> {WHITE_SPACE}+                            { yybegin(WAITING_VERSION); return TokenType.WHITE_SPACE; }
+<WAITING_NAME> {KEY_CHARACTER}                   { yybegin(WAITING_VERSION); return VersionPropsTypes.NAME_KEY; }
```
crogoz: nit - maybe change this to `KEY` (as it can contain multiple characters)

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
```diff
@@ -23,52 +23,52 @@ import com.intellij.psi.TokenType;
...
 
-<WAITING_VALUE> {VALUE_CHARACTER}+                          { yybegin(INVALID_VALUE); return VersionPropsTypes.VERSION; }
+<WAITING_COMMENT> {WHITE_SPACE}*{COMMENT}        { yybegin(INVALID_VALUE); return VersionPropsTypes.COMMENT; }
+<WAITING_COMMENT> [^\n]+                         { return TokenType.BAD_CHARACTER; }
 
-<INVALID_VALUE> [^\n]+                                      { return TokenType.BAD_CHARACTER; }
+<INVALID_VALUE> [^\n]+                           { return TokenType.BAD_CHARACTER; }
 
-({CRLF}|{WHITE_SPACE})+                                     { yybegin(YYINITIAL); return TokenType.WHITE_SPACE; }
+({CRLF}|{WHITE_SPACE})+                          { yybegin(YYINITIAL); return TokenType.WHITE_SPACE; }
```
crogoz: I think this will allow for the start of the line to be a white space: 
` com.palantir....=1.0.0`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
FinlayRJW: done

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
FinlayRJW: good spot! - Ive fixed it now but still allowed white space before comments

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
```diff
@@ -23,52 +23,50 @@ import com.intellij.psi.TokenType;
...
-<INVALID_VALUE> [^\n]+                                      { return TokenType.BAD_CHARACTER; }
+<WAITING_COMMENT> {WHITE_SPACE}*{COMMENT}        { yybegin(INVALID_VALUE); return VersionPropsTypes.COMMENT; }
+<WAITING_COMMENT> [^\n]+                         { return TokenType.BAD_CHARACTER; }
 
-({CRLF}|{WHITE_SPACE})+                                     { yybegin(YYINITIAL); return TokenType.WHITE_SPACE; }
+{CRLF}+                                          { yybegin(YYINITIAL); return TokenType.WHITE_SPACE; }
 
-[^]                                                         { return TokenType.BAD_CHARACTER; }
\ No newline at end of file
+[^]                                              { return TokenType.BAD_CHARACTER; }
```
CRogers, line 72: Is there any way to write lexer tests? Like give it a string and see if it produces some set of tokens?

FinlayRJW: So I did have a very basic test that was testing that the tokens matched but it was very simplified. I have now added a test for a much more complex token string with all the edge cases I could think of

FinlayRJW: @CRogers are you happy with the new tests?

gradle-consistent-versions-idea-plugin/src/test/resources/lexerTests/24bad4b06b90bba33694afd3e19b04a056cdbf8098e1eca917b90661a5999666
```diff
@@ -0,0 +1,18 @@
+VersionPropsTokenType.GROUP_PART has
```
CRogers, line 2: Can you put the test input at the top of the file? Otherwise it's hard to see what the input was.

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsLexerTests.java
```diff
@@ -0,0 +1,167 @@
...
+    private static Path getOutputFilePath(String input) throws IOException {
+        String fileName = hashToFolderName(input);
+        Path outputPath = Paths.get("src/test/resources/lexerTests/" + fileName);
+        Files.createDirectories(outputPath.getParent());
+        return outputPath;
+    }
+
+    private static String hashToFolderName(String input) {
+        try {
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
```
CRogers: You can use Guava here: `Hashing.sha256().hashString(input, StandardCharsets.UTF_8).toString()`

gradle-consistent-versions-idea-plugin/src/test/resources/lexerTests/24bad4b06b90bba33694afd3e19b04a056cdbf8098e1eca917b90661a5999666
FinlayRJW: done :)

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsLexerTests.java
```diff
@@ -0,0 +1,156 @@
...
+                "odd/.characters@?'.in!:all\\\\/ = the_pl4ces.,?",
+                "has.end.of:line=comments # here is the comment",
+                " has.extra.space:bad_token = 1",
+                "has . space . around:dots = 1",
+                "# A comment",
+                "has.spaces.around : colon = 1",
+                "has.no.spaces.around:equals=1",
+                "has.loads.of.spaces:around   =   equals",
+                "    # comment after spaces"
+            })
```
CRogers: There's one part which is missing: if you were to change one of these tests, you would never automatically delete the existing test. I think we can do this though: change this `@ValueSource` into a `@MethodSource`, have a method that returns all these test cases. Then make one `@Test` in this file that:

1. Reads all the test cases from the method
1. If on CI: errors if there are any extra tests in the `lexerTests` directory, telling people to run the test locally to fix (can use `describedAs` in assertj).
2. If not on CI: just deletes the extra tests in the `lexerTests` directory

FinlayRJW: so I think I have done this, now a set of tests is passed in and we check if the directory is in the expected form

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsLexerTests.java
```diff
@@ -90,27 +69,78 @@ private static void deleteDirectoryRecursively(Path path) throws IOException {
...
                 "has.loads.of.spaces:around   =   equals",
-                "    # comment after spaces"
-            })
+                "    # comment after spaces"));
+    }
+
     @ParameterizedTest
-    public void test_psi_tree_structure(String input) throws Exception {
+    @MethodSource("provideTestCases")
+    public void test_psi_tree_structure(Set<String> input) throws Exception {
```
CRogers: If it's taking a `Set<String>`, at this point there's not point having this be a `@ParameterisedTest` as it only takes one input (I was suggesting have another, separate `@Test` to check to for tests that need deleting). This is also fine too, just turn this `@ParameterisedTest` into a plain `@Test` and have the test cases defined in the test body.

CRogers:
👍🏻

PR #22, opened 2024-10-01
fix missing import

FinlayRJW:
## Before this PR
All builds where breaking

## After this PR
Fixed builds by importing `import org.junit.jupiter.api.Assertions;`

Not sure how a breaking change like that managed to merge but this should fix it


FinlayRJW:
was able to fix this in the excavator

PR #23, opened 2024-10-01
use gradle cache

FinlayRJW:
## Before this PR
To show auto complete we have to visit the remote repo, for some repositories this can be extremely slow.

## After this PR
We can now use the local gradle cache to do auto complete for name and group. I opted not todo this for version as it changes at a much higher frequency.

This means the user gets auto suggestions nearly instantly.

The complexity with this change is that we dont know where packages in the gradle cache originated from so a particular project may not have access to a certain suggested package. To combat this the suggestions from the cache are labeled and will be overwritten by the remote suggestions when they have loaded. 

## Possible downsides?
It might be a bit confusing to see some suggestions labeled as from cache


CRogers:
I think the changelog (and also PR title) can be improved here. Something like "Use Gradle cache to source `versions.props` completions for better performance"

FinlayRJW:
So I think this should be in a good state now - it still has no way of distinguishing between packages from different repos from the cache but I don't think thats to much of a concern

FinlayRJW:
This has been superseded by https://github.com/palantir/gradle-consistent-versions-idea-plugin/pull/35 which reads the bin files to get all the information we need

PR #25, opened 2024-10-02
ignore unknown

FinlayRJW:
## Before this PR
We would silently fail to parse metadata for versions as sometime the metadata contains extra fields

## After this PR
we know ignore unknown fields so that it does not fail in this way - I have also made the message error level and not debug so it is no longer silently failing

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Versioning.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Versioning.java
@@ -16,6 +16,7 @@
 
 package com.palantir.gradle.versions.intellij;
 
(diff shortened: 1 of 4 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Versioning.java
```diff
@@ -25,6 +26,7 @@
 @Value.Immutable
 @JsonDeserialize(as = ImmutableVersioning.class)
 @JsonSerialize(as = ImmutableVersioning.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
CRogers, line 29: The sneakiest most common problem with jackson

CRogers (approved):

PR #27, opened 2024-10-03
Settings Page

FinlayRJW:
## Before this PR
No way to enable the plugin running wVL on a project by project level

## After this PR
We now have a project level setting under the `tools` section that allows you to disable the feature

## Possible downsides?
Might result in users turning off the feature then not switching it back on

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsProjectSettings.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsProjectSettings.java
@@ -18,43 +18,26 @@ package com.palantir.gradle.versions.intellij;
 
 import com.intellij.openapi.components.PersistentStateComponent;
 import com.intellij.openapi.components.Service;
+import com.intellij.openapi.components.Service.Level;
 import com.intellij.openapi.components.Storage;
 import com.intellij.openapi.project.Project;
 import org.jetbrains.annotations.Nullable;
 
+@Service(Level.PROJECT)
 @com.intellij.openapi.components.State(
         name = "ProjectSettings",
         storages = {@Storage("gcv-plugin-settings.xml")})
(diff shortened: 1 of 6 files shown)


FinlayRJW:
Now we are no long including the list of repos in the settings so this is purely to allow user to set wether the wVL runs on a project by project level

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsProjectSettings.java
```diff
@@ -0,0 +1,89 @@
...
+    public enum EnabledState {
+        UNKNOWN,
+        ENABLED,
+        DISABLED
+    }
+
+    public static final class State {
+        // Using an enum like this ensure that the file is created in .idea allowing for end users to configure it for
+        // all users
+        private EnabledState enabled = EnabledState.UNKNOWN;
```
CRogers: We were going to get rid of this, right?

FinlayRJW: so I left this in as without it intillij just doesnt seem to create a `gcv-plugin-settings.xml` the setting still works exactly as you would expect even if `gcv-plugin-settings.xml` isnt generated but I was worried it wouldnt give repo owners a way of disabling the setting for all users

FinlayRJW: Okay I found a happy middle ground - if I make the state as below then when enabled is false a file is created - this also means that a user can manually create the file to set enabled to false
```
    public static final class State {
        private boolean enabled = true;

        public void setEnabled(boolean enabled) {
            this.enabled = enabled;
        }

        public boolean getEnabled() {
            return this.enabled;
        }
    }
```

CRogers:
:+1:

PR #31, opened 2024-10-07
Read repositories on a per project level

FinlayRJW:
## Before this PR
We had been using a hard coded default repo of maven central rather than actually reading in a projects maven repos 

## After this PR
Now we read the repositories from the `gcv-maven-repositories.xml` file, this also required a change to how the packages where cached in memory such that the repo url is inlcuded 
==COMMIT_MSG==
Read repositories on a per project level
==COMMIT_MSG==

## Possible downsides?
I have it set so if no repos are found in a project it still uses maven central as a default. There is no threading so if a very slow repo is the first in a list it will cause all suggestions to be queued behind it


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
```diff
@@ -43,11 +43,11 @@ private void extendCompletion(IElementType elementType) {
...
 
-                List<String> repositories = List.of("https://repo1.maven.org/maven2/");
-
                 DependencyGroup group = DependencyGroup.groupFromParameters(parameters);
 
-                repositories.stream()
+                Project project = parameters.getOriginalFile().getProject();
+
+                RepositoryLoader.loadRepositories(project).stream()
                         .map(RepositoryExplorer::new)
```
felixdesouza: similar thing, your `RepositoryExplorer` contains a cache, but you're not sharing them, between repositories and also between `ElementType`s despite the contents of what's in the cache being the same

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -37,7 +38,7 @@ public class RepositoryExplorer {
     private static final Logger log = LoggerFactory.getLogger(RepositoryExplorer.class);
 
     private final String baseUrl;
-    private static final Cache<DependencyGroup, List<Folder>> folderCache = Caffeine.newBuilder()
+    private static final Cache<CacheKey, List<Folder>> folderCache = Caffeine.newBuilder()
             .expireAfterWrite(10, TimeUnit.MINUTES)
             .maximumSize(100)
             .build();
```
felixdesouza: hmm, it's not clear to me why you need to have a separate cache key type if this is never going to contain a different base url?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
FinlayRJW: So what was happening here was that I had the caffine cache within the `RepositoryExplorer` as a static, this worked but made it hard to understand, to fix this now `repositoryExplorer` is static within the context of the completion contributor and the url is passed into the methods for getting the completions

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
FinlayRJW: So I think this is needed now that `repositoryExplorer` is a field in the completion contributor and the cache is correctly shared - we need to have a cache of the folders at a particular depth for each URL, this deals with situations where multiple urls have the same groups but different child folders.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryLoader.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final ObjectMapper XML_MAPPER = new XmlMapper().registerModule(new GuavaModule());
+    private static final String MAVEN_REPOSITORIES_FILE_NAME = ".idea/gcv-maven-repositories.xml";
+    private static final String DEFAULT = "https://repo.maven.apache.org/maven2/";
+
+    private RepositoryLoader() {
+        // Utility class; prevent instantiation
+    }
+
+    public static Set<String> loadRepositories(Project project) {
+        Set<String> urls = new HashSet<>();
```
CRogers: This having a single variable that's used in a bunch of places is a bit confusing. Instead, I'd return `Set.of(DEFAULT)` when it's just the default.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryLoader.java
```diff
@@ -0,0 +1,94 @@
...
+        if (!mavenRepoFile.exists()) {
+            urls.add(DEFAULT);
+            return urls;
+        }
+
+        try {
+            Repositories repositories = XML_MAPPER.readValue(mavenRepoFile, Repositories.class);
+            for (RepositoryConfig config : repositories.repositories()) {
+                urls.add(config.url());
+            }
```
CRogers: And here use streams so:

```java
repositories.repositories().stream().map(RepositoryConfig::url).collect(Collectors.toSet())
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryLoader.java
```diff
@@ -0,0 +1,94 @@
...
+            urls.add(DEFAULT);
+        }
+
+        return urls;
+    }
+
+    @Value.Immutable
+    @JsonDeserialize(as = ImmutableRepositoryConfig.class)
+    @JsonSerialize(as = ImmutableRepositoryConfig.class)
+    interface RepositoryConfig {
```
CRogers, line 66: Are these type defined in both this repo and the main GCV repo? If so, it would have probably made sense to have them shared in some way, ie make a new project in this repo then have the GCV plugin use those types. Fine for now, but we should definitely share them if we end up making a change.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -36,44 +37,41 @@
...
 
-    public RepositoryExplorer(String baseUrl) {
-        this.baseUrl = baseUrl;
-    }
-
-    public final List<Folder> getFolders(DependencyGroup group) {
-        List<Folder> folders = folderCache.get(group, key -> {
-            List<Folder> loadedFolders = loadFolders(key);
+    public final Set<Folder> getFolders(DependencyGroup group, String url) {
```
CRogers: I really think "folder" is not a good name for this (maven doesn't have a concept of "folders"). What does it actually represent?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryLoader.java
FinlayRJW: yeah this is identical to what is in GCV, I've created an issue on this repo to track this

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
FinlayRJW: It represents one folder in the remote repo, so for example `com` would be a folder and so would `palantir`. so `getFolders` returns a list of all folders based on the current group (which might only be part of a group and not the whole thing) - honestly not sure if it really needs to be a type could probably just use `String`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryLoader.java
```diff
@@ -0,0 +1,88 @@
...
+
+public final class RepositoryLoader {
+    private static final Logger log = LoggerFactory.getLogger(RepositoryExplorer.class);
+
+    private static final ObjectMapper XML_MAPPER = new XmlMapper().registerModule(new GuavaModule());
+    private static final String MAVEN_REPOSITORIES_FILE_NAME = ".idea/gcv-maven-repositories.xml";
+    private static final String DEFAULT = "https://repo.maven.apache.org/maven2/";
+
+    private RepositoryLoader() {
+        // Utility class; prevent instantiation
```
CRogers: There's generally no need to comment this, it's a very common pattern. Also, for these utility classes I think it's best to just put the constructor at the bottom to show how much less it matters.

FinlayRJW: thanks have also updated `ContentsUtil`

FinlayRJW:
I've added a `ExternalSystemTaskNotificationListener` to listen for gradle refreshed and invalidate the `GradleCacheExplorer`'s cache. This should make it safe to pass the URLs into the `getCompletions` method whilst still caching on a static key as the cache is invalidated anytime that the URLs could change - I think this should address the concerns from https://github.com/palantir/gradle-consistent-versions-idea-plugin/pull/35#discussion_r1793640770

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleProjectRefreshListener.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.intellij.openapi.externalSystem.model.task.ExternalSystemTaskId;
+import com.intellij.openapi.externalSystem.model.task.ExternalSystemTaskNotificationEvent;
+import com.intellij.openapi.externalSystem.model.task.ExternalSystemTaskNotificationListener;
+import com.intellij.openapi.externalSystem.model.task.ExternalSystemTaskType;
+import org.jetbrains.plugins.gradle.util.GradleConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GradleProjectRefreshListener implements ExternalSystemTaskNotificationListener {
```
CRogers: I would probably call this something more specific to the cache invalidation, something like `InvalidateCacheOnGradleProjectRefresh`? We might end up with more than one of these, and would be good to keep them separate rather than putting all the logic into one listener.

CRogers:
:+1: Given comment

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
FinlayRJW: have given a more verbose name of `GroupPartOrPackageName`

CRogers:
:+1:

PR #32, opened 2024-10-07
Create file system cache

FinlayRJW:
## Before this PR
We had no long term cache so between runs we would lose any information gathered from the remote repos 

## After this PR
Now the plugin builds a user by user local cache similar to the gradle cache, this is better than just using the gradle cache as we have full control and can mark which repos packages are associated with 
==COMMIT_MSG==
Create file system cache
==COMMIT_MSG==

## Possible downsides?
The plugin now creates this cache which could become very large over time.

So that we can determine if a package is complete (i.e. add the ':') we have to make a lot of calls to check if `maven-metadata` exists, this is only done once per completion but still this might be a bit hard on the remote repos. Annoyingly packages are not always on the same level so we cannot just check the first folder.

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryFileCache.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryFileCache.java
@@ -16,15 +16,15 @@
 
 package com.palantir.gradle.versions.intellij;
 
+import com.google.common.hash.Hashing;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -46,7 +46,7 @@ public class RepositoryFileCache {
     private final ExecutorService executorService = Executors.newFixedThreadPool(20);
 
     public final void syncCache(String repoUrl, Set<String> packages) {
-        CompletableFuture.runAsync(
+        CompletableFuture<Void> unused = CompletableFuture.runAsync(
                 () -> {
                     Set<String> existingPackages = cache.get(repoUrl);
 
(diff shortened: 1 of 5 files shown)


FinlayRJW:
@CRogers  I'll be interest in what you think of this - essentially I wasn't happy with using the gradle cache as we have no control of it and cannot easily determine a packages origin repo - this now lets us build our cache as people use the plugin, it means we don't get that instant win of using people existing gradle cache but I think the extra control is probably worth it. I'm a bit dubious of checking if a package is complete as the calls are quite expensive but it allows for full autocompletes rather than only completing section by section

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryFileCache.java
```diff
@@ -0,0 +1,160 @@
...
+public class RepositoryFileCache {
+    private static final Logger log = LoggerFactory.getLogger(RepositoryFileCache.class);
+
+    private static final String CACHE_PATH = System.getProperty("user.home") + "/.gcv-cache";
+
+    private static final Map<String, Set<String>> cache = new HashMap<>();
+
+    private final ExecutorService executorService = Executors.newFixedThreadPool(20);
+
+    public final void syncCache(String repoUrl, Set<String> packages) {
```
CRogers, line 48: Writing a good cache well is actually really tricky. Probably best if we review this in person so I can explain. But the trick is to structure it in a way where you can just slot the cache as another layer rather than all in one class (ie `new CachedThing(new UncachedThing())`)

CRogers: It's also really tricky to do this properly in the presence of concurrent readers/writers.

FinlayRJW: Interesting, I did have a look around and couldn't find any good packages todo this form me, in general do you think using a file cache rather than / as well as the gradle cache is worth it?

CRogers: Honestly, in this situation I'd consider using something like sqlite to make the cache as that will handle all the hard bits for you. But I would also question what this kind of caching will bring:
1. We'll still need to check every so often for new packages in case something has been newly added.
2. It's only going to cover all the cases that have been individually searched via the idea plugin rather than the Gradle cache just having everything ever used in Gradle builds.

FinlayRJW: good point, I was trying to avoid the issue with gradle cache where we don't know the source of the package but I'm not sure how much of a real problem that is - given we mark the suggestions from the cache clearly

PR #35, opened 2024-10-08
Read gradle cache bin files to get suggetions

FinlayRJW:
## Before this PR
Previously two different methods of caching have been suggested, the first was to use the gradle cache files2.1 to get the suggestions this was however had no way to know where the packages came from. The second was to build our own cache although this was complex and would have limited utility

## After this PR
Now we directly parse the bin files gradle uses to extract suggestions and match them to the projects repos, this is the best of both worlds as we filter by repo and get all the advantages of using the gradle cache
==COMMIT_MSG==
Read gradle cache bin files to get suggetions
==COMMIT_MSG==

## Possible downsides?
We are parsing a bin file, this could cause issues in the future - however as we are only giving suggestions if we match the repo url this should mitigate the chance of gradle changes breaking the plugin


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
```diff
@@ -56,6 +58,25 @@ protected void addCompletions(
         });
     }
 
+    private void cacheCompletion(IElementType elementType) {
+        extend(CompletionType.BASIC, PlatformPatterns.psiElement(elementType), new CompletionProvider<>() {
+            @Override
+            protected void addCompletions(
+                    CompletionParameters parameters, ProcessingContext context, CompletionResultSet resultSet) {
+
+                List<String> repositories = List.of("https://repo.maven.apache.org/maven2/");
```
felixdesouza: I imagine this is temporary?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+                result.add(currentString.toString());
+            }
+        } catch (IOException e) {
+            log.error("Failed to extract strings from bin file", e);
+        }
+
+        return result;
+    }
+
+    final boolean isPrintableChar(char ch) {
```
felixdesouza: you should use

    !Character.isISOControl(ch)

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+                    .collect(Collectors.toSet());
+        } catch (RuntimeException e) {
+            log.debug("Operation was cancelled", e);
+        } catch (Exception e) {
+            log.warn("Failed to get completions", e);
+        }
+        return new HashSet<>();
+    }
+
+    private Callable<Set<String>> extractStrings(ProgressIndicator indicator) {
```
felixdesouza: nit, have this return a `Set<String>>` and do the callable lambda stuff at the callsite

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+        return () -> cache.get("metadata", key -> {
+            Set<String> stringsSet = new HashSet<>();
+            File gradleCacheFolder = new File(GRADLE_CACHE_PATH);
+            File[] metadataFolders = gradleCacheFolder.listFiles((dir, name) -> name.startsWith("metadata-"));
+
+            if (metadataFolders != null) {
```
felixdesouza: Generally, I prefer to reduce nesting where possible as it reduces the amount of states one needs to track mentally.

```java
File gradleCacheFolder = new File(GRADLE_CACHE_PATH);
File[] metadataFolders = gradleCacheFolder.listFiles((dir, name) -> name.startsWith("metadata-"));

if (metadataFolders == null) {
  return Set.of();
}

for (File metadataFolder : metadataFolders) {
  // ...
}
```

felixdesouza: Also prefer to use `Files` because you get a stream back directly.

```java
try (Stream<Path> stream = Files.list(GRADLE_CACHE_PATH)) {
  stream.filter(path -> path.getFileName().toString().startsWith("metadata-"))
    // you could probably even make this a helper method to make it clear what you're doing
    .map(metadataFolder -> metadataFolder.resolve("resource-at-url.bin"))
    .flatMap(GradleCacheExplorer::extractStringsFromBinFile)
}
```

You then have something like:

```
final Stream<String> extractStringsFromBinFile(Path binFile)  {
  if (Files.notExists(binFile)) {
    return Stream.of();
  }

  // whatever else you gotta do here
}
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+        try (BufferedInputStream bis = new BufferedInputStream(new FileInputStream(binFile))) {
+            StringBuilder currentString = new StringBuilder();
+            int byteValue;
+
+            while ((byteValue = bis.read()) != -1) {
+                char charValue = (char) byteValue;
+                if (isPrintableChar(charValue)) {
+                    currentString.append(charValue);
+                } else {
+                    if (currentString.length() >= 4) {
```
felixdesouza: not sure I understand the significance of the string being `>= 4` ?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        return result;
+    }
+
+    final boolean isPrintableChar(char ch) {
+        return ch >= 32 && ch <= 126; // Printable ASCII range
+    }
+
+    final String sanitiseUrl(String url) {
```
felixdesouza: I don't think this name indicates what it's trying to do?

Looking at the tests, it extracts the group from the url in the form of 

    com/example/artifact/1.0/artifact-1.0.jar -> com.example:artifact

So this seems to be a: `extractGroupAndArtifactFromUrl` method

I would javadoc this and include an example.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+        String finalUrl = url;
+        for (String projectUrl : projectUrls) {
+            if (finalUrl.startsWith(projectUrl)) {
+                finalUrl = finalUrl.substring(projectUrl.length());
+                break;
```
felixdesouza: this is also confusing to me, this feels like it will spit out garbage if no urls match right?

why not have this method return `Optional<String>` and in this case, return `Optional.empty()` when we don't haven't managed to match the url with any project urls.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+
+        int lastSlashIndex = finalUrl.lastIndexOf('/');
+        int secondLastSlashIndex = finalUrl.lastIndexOf('/', lastSlashIndex - 1);
+
+        if (secondLastSlashIndex != -1) {
+            finalUrl = finalUrl.substring(0, secondLastSlashIndex).replace('/', '.');
+            finalUrl = finalUrl.replaceFirst("\\.([^.]*)$", ":$1");
+        }
```
felixdesouza: similarly here, if we don't find a second last slash index, the string returned is garbage no?

felixdesouza: so I would do something similar here and return an `Optional.empty`

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -0,0 +1,86 @@
...
+        explorer = new GradleCacheExplorer(projectUrls);
+    }
+
+    @Test
+    void test_gets_valid_urls_only() {
+        assertTrue(explorer.isValidResourceUrl(
+                "https://repo.maven.apache.org/maven2/com/example/artifact/1.0/artifact-1.0.pom"));
+        assertTrue(
+                explorer.isValidResourceUrl("https://jcenter.bintray.com/com/example/artifact/1.0/artifact-1.0.jar"));
+        assertFalse(explorer.isValidResourceUrl("https://example.com/com/example/artifact/1.0/artifact-1.0.pom"));
```
felixdesouza: so I actually quite like using `assertJ` for these assertions because you can include a reason as to why something should be asserted the way the way it is.

Just looking at this, I'm not actually sure what I'm supposed to be looking for here.

An example assertion with assertj would look like:

```java
assertThat(explorer.isValidResourceUrl(...))
  .as("because it includes some character") // I actually do not know
  .describedAs(..) // is an alternative but equivalent usage, I prefer `as`
  .isFalse()
```

It's a bit more verbose, but it's clearer what you're actually trying to test

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertFalse(explorer.isValidResourceUrl(
+                "https://repo.maven.apache.org/maven2/com/example/artifact/1.0/artifact-1.0.txt"));
+    }
+
+    @Test
+    void test_gets_all_strings_from_bin() throws IOException {
+        File tempFile = File.createTempFile("test", ".bin");
+        tempFile.deleteOnExit();
```
felixdesouza: You should just use [`@TempDir`](https://www.baeldung.com/junit-5-temporary-directory#tmpdir-extension-1)

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -0,0 +1,86 @@
...
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(tempFile, StandardCharsets.UTF_8))) {
+            writer.write("hello.jar\nworld.pom\nanother.jar\b\f");
+        }
+
+        Set<String> result = explorer.extractStringsFromBinFile(tempFile);
+        Set<String> expected = new HashSet<>(List.of("hello.jar", "world.pom", "another.jar"));
+
+        assertEquals(expected, result);
```
felixdesouza: with assertj you'd have something like:

```java
assertThat(explorer.extractStringsFromBinFile(...))
  .containsOnly("hello.jar", "world.pom", "another.jar");
```

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @Test
+    void test_gets_all_strings_from_bin() throws IOException {
+        File tempFile = File.createTempFile("test", ".bin");
+        tempFile.deleteOnExit();
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(tempFile, StandardCharsets.UTF_8))) {
+            writer.write("hello.jar\nworld.pom\nanother.jar\b\f");
+        }
+
+        Set<String> result = explorer.extractStringsFromBinFile(tempFile);
+        Set<String> expected = new HashSet<>(List.of("hello.jar", "world.pom", "another.jar"));
```
felixdesouza: invalidated by my assertj comment, but could you not just do `Set.of(...)` ?

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertEquals(expected, result);
+    }
+
+    @Test
+    void test_is_printable_char() {
+        assertTrue(explorer.isPrintableChar('A'));
+        assertTrue(explorer.isPrintableChar(' '));
+        assertFalse(explorer.isPrintableChar('\n'));
+        assertFalse(explorer.isPrintableChar((char) 127));
+    }
```
felixdesouza: I think this largely goes away if you just use the method from `Character`. Although you could also keep the test to make sure it does what you want

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Test
+    void test_url_sanitise_correctly() {
```
felixdesouza: change the name of this to match the new name of the method

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+            return new HashSet<>();
+        }
+
+        try {
+            Future<Set<String>> future =
+                    ApplicationManager.getApplication().executeOnPooledThread(extractStrings(indicator));
+            Set<String> results =
+                    com.intellij.openapi.application.ex.ApplicationUtil.runWithCheckCanceled(future::get, indicator);
+
+            return results.stream()
```
felixdesouza, line 73: why don't we short-circuit here and do:

```java
if (parsedInput.isEmpty()) {
  return results;
}
```

I believe they are equivalent, but you're not just constructing the set over again

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public GradleCacheExplorer(List<String> projectUrls) {
+        this.projectUrls = projectUrls;
+    }
+
+    public final Set<String> getCompletions(DependencyGroup input) {
+        String parsedInput = String.join(".", input.parts());
+        ProgressIndicator indicator = ProgressManager.getInstance().getProgressIndicator();
+        if (indicator == null) {
+            return new HashSet<>();
```
felixdesouza: you should return a `Set.of()` or `Collections.emptySet()`. It returns a specialised implementation of `Set` without all of the baggage associated with a `HashSet` (which creates a `HashMap`). 

This applies for all the other places where you are returning an empty set as some sort of bailing out mechanism

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
felixdesouza: whatever the reason, I think it's worth commenting or extracting the `4` into a constant that explains your intent? e.g. `PROBABLE_WORD_MINIMUM_LENGTH` idk

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+    final Set<String> extractStringsFromBinFile(File binFile) {
+        Set<String> result = new HashSet<>();
+        try (BufferedInputStream bis = new BufferedInputStream(new FileInputStream(binFile))) {
+            StringBuilder currentString = new StringBuilder();
+            int byteValue;
+
+            while ((byteValue = bis.read()) != -1) {
+                char charValue = (char) byteValue;
+                if (isPrintableChar(charValue)) {
+                    currentString.append(charValue);
```
felixdesouza, line 132: nit: can you add a `continue` and then drop the `else` block?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
```diff
@@ -56,6 +58,25 @@ protected void addCompletions(
...
+
+                DependencyGroup group = DependencyGroup.groupFromParameters(parameters);
+                GradleCacheExplorer gradleCacheExplorer = new GradleCacheExplorer(repositories);
```
felixdesouza: what is the lifecycle of the enclosing method? is `addCompletions` going to be called on every completion suggestion? if so you'll end up gc-ing a lot and making a bunch of use once only `GradleCacheExplorer` objects. 

My gut feeling is that you *may* want `gradleCacheExplorer` to be a field in `FolderCompletionContributor`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
```diff
@@ -32,18 +32,20 @@
 public class FolderCompletionContributor extends CompletionContributor {
 
     public FolderCompletionContributor() {
-        extendCompletion(VersionPropsTypes.GROUP_PART);
-
-        extendCompletion(VersionPropsTypes.NAME_KEY);
+        IElementType[] elementTypes = {VersionPropsTypes.GROUP_PART, VersionPropsTypes.NAME_KEY};
+        for (IElementType elementType : elementTypes) {
+            cacheCompletion(elementType);
```
felixdesouza: yeah, each element type is not going to share a cache here.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
```diff
@@ -32,18 +32,20 @@
 public class FolderCompletionContributor extends CompletionContributor {
 
     public FolderCompletionContributor() {
-        extendCompletion(VersionPropsTypes.GROUP_PART);
-
-        extendCompletion(VersionPropsTypes.NAME_KEY);
+        IElementType[] elementTypes = {VersionPropsTypes.GROUP_PART, VersionPropsTypes.NAME_KEY};
```
felixdesouza: I would just inline this:

```java
cacheCompletion(VersionPropsTypes.GROUP_PART);
cacheCompletion(VersionPropsTypes.NAME_KEY);
remoteCompletion(VersionPropsTypes.GROUP_PART);
remoteCompletion(VersionPropsTypes.NAME_KEY);
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,155 @@
...
+                    ApplicationManager.getApplication().executeOnPooledThread(extractStrings(indicator));
+            Set<String> results =
+                    com.intellij.openapi.application.ex.ApplicationUtil.runWithCheckCanceled(future::get, indicator);
+
+            return results.stream()
+                    .filter(result -> result.startsWith(parsedInput))
+                    .map(result -> parsedInput.isEmpty() ? result : result.substring(parsedInput.length() + 1))
+                    .collect(Collectors.toSet());
+        } catch (RuntimeException e) {
+            log.debug("Operation was cancelled", e);
```
felixdesouza, line 78: is this accurate? you any other `RuntimeException` could be thrown as a result of doing something other than observing the cancellation from the `Indicator`.

felixdesouza (dismissed):

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
FinlayRJW: Yep this will be replaced by this https://github.com/palantir/gradle-consistent-versions-idea-plugin/pull/31

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
FinlayRJW: Maybe I not understood something but I thought that because the caffine cache is static it would be share across instances 

`private static final Cache<String, Set<String>> cache`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: I've updated this to use a Files so that it can all be a stream I think this is much cleaner is this what you meant?

``` java
    private Set<String> extractStrings(ProgressIndicator indicator) {
        return cache.get("metadata", key -> {
            try (Stream<Path> metadataFolders = Files.list(Paths.get(GRADLE_CACHE_PATH))
                    .filter(path -> path.getFileName().toString().startsWith("metadata-"))) {
                return metadataFolders
                        .filter(metadataFolder -> !indicator.isCanceled())
                        .map(metadataFolder -> metadataFolder.resolve("resource-at-url.bin"))
                        .filter(Files::exists)
                        .flatMap(this::extractStringsFromBinFile)
                        .filter(this::isValidResourceUrl)
                        .map(this::extractGroupAndArtifactFromUrl)
                        .flatMap(Optional::stream)
                        .collect(Collectors.toSet());
            } catch (IOException e) {
                log.error("Failed to list metadata folders", e);
                return Collections.emptySet();
            }
        });
    }
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: Yeah 4 was just an arbitrary choice probably better just to use `!currentString.isEmpty()`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: Have updated this with a docstring and changed the name

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: Have updated this now to use optional so should be better now:

``` java
    Optional<String> extractGroupAndArtifactFromUrl(String url) {
        return projectUrls.stream().filter(url::startsWith).findFirst().flatMap(projectUrl -> {
            String finalUrl = url.substring(projectUrl.length());
            int lastSlashIndex = finalUrl.lastIndexOf('/');
            int secondLastSlashIndex = finalUrl.lastIndexOf('/', lastSlashIndex - 1);

            if (lastSlashIndex == -1 || secondLastSlashIndex == -1) {
                return Optional.empty();
            }

            finalUrl = finalUrl.substring(0, secondLastSlashIndex).replace('/', '.');
            finalUrl = finalUrl.replaceFirst("\\.([^.]*)$", ":$1");
            return Optional.of(finalUrl);
        });
    }
```

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
FinlayRJW: do we generally prefer assertj?

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
FinlayRJW: yeah dont need to test this anymore

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: good to know thanks!

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: no this was not accurate I believe we need to catch `ProcessCanceledException ` and not RuntimeException

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
FinlayRJW: So I had the actual caffine cache as static in the object but this makes it very unclear so I have now made `gradleCacheExplorer` a field in `FolderCompletionContributor` which reads much better

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: After our chat yesterday I have updated this to throw the exception properly

FinlayRJW (commented):
Thanks so much for such a detailed review all changes have been implemented

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
```diff
@@ -31,19 +31,23 @@
 
 public class FolderCompletionContributor extends CompletionContributor {
 
-    public FolderCompletionContributor() {
-        extendCompletion(VersionPropsTypes.GROUP_PART);
+    private static final GradleCacheExplorer gradleCacheExplorer =
```
felixdesouza: this can actually just be an instance type, assuming you only have one instance of `FolderCompletionContributor`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,175 @@
...
+        } catch (Exception e) {
+            log.warn("Failed to get completions", e);
+        }
+        return Collections.emptySet();
+    }
+
+    private Set<String> extractStrings(ProgressIndicator indicator) {
+        return cache.get("metadata", key -> {
+            try (Stream<Path> metadataFolders = Files.list(Paths.get(GRADLE_CACHE_PATH))
+                    .filter(path -> path.getFileName().toString().startsWith("metadata-"))) {
```
felixdesouza: nit: can you move this filter outside of the `try-with-resources` initialiser? makes it a bit hard to read

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,175 @@
...
+     * <p>Example: For the URL "http://example.com/org/example/project/1.0/project-1.0.jar"
+     * and "http://example.com/" in the list of projectUrls, it returns "org.example.project:project".
+     *
+     * @param url the URL to process
+     * @return an {@link Optional} containing a string in the format "group:artifact" if extraction is successful,
+     *         or {@link Optional#empty()} if no matching project URL is found or the URL does not have the expected structure.
+     */
+    Optional<String> extractGroupAndArtifactFromUrl(String url) {
+        return projectUrls.stream().filter(url::startsWith).findFirst().flatMap(projectUrl -> {
+            String finalUrl = url.substring(projectUrl.length());
```
felixdesouza: let's call this `mavenLayout` as per https://maven.apache.org/repositories/layout.html

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,175 @@
...
+                result.add(currentString.toString());
+            }
+        } catch (IOException e) {
+            log.error("Failed to extract strings from bin file", e);
+        }
+        return result.stream();
+    }
+
+    /**
+     * Extracts the group and artifact identifiers from a given URL.
```
felixdesouza:
```suggestion
     * Extracts the group and artifact identifiers from a given maven2 layout URL.
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -0,0 +1,175 @@
...
+            String finalUrl = url.substring(projectUrl.length());
+            int lastSlashIndex = finalUrl.lastIndexOf('/');
+            int secondLastSlashIndex = finalUrl.lastIndexOf('/', lastSlashIndex - 1);
+
+            if (lastSlashIndex == -1 || secondLastSlashIndex == -1) {
+                return Optional.empty();
+            }
+
+            finalUrl = finalUrl.substring(0, secondLastSlashIndex).replace('/', '.');
+            finalUrl = finalUrl.replaceFirst("\\.([^.]*)$", ":$1");
```
felixdesouza: I'm still struggling to parse this a little. A bit more verbose, but could we do:

```java
int thirdLastSlashIndex = ...

String group = mavenLayout.substring(0, thirdLastSlashIndex).replace('/', '.');
String artifact = mavenLayout.substring(thirdLastSlashIndex, secondLastSlashIndex);
return String.format("%s:%s", group, artifact);
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
felixdesouza: otherwise you've kinda shifted the problem around a bit. Ideally we only have mutable state as instance fields where possible. Sometimes this is unavoidable e.g. `ThreadLocal`s or where we don't control initialisation, but this doesn't seem to be the case here.

felixdesouza (dismissed):
approved pending changes

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/FolderCompletionContributor.java
FinlayRJW: yep should only be one instance

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -82,10 +75,13 @@ public final Set<String> getCompletions(DependencyGroup input) {
         return Collections.emptySet();
     }
 
-    private Set<String> extractStrings(ProgressIndicator indicator) {
+    private Set<String> extractStrings(Set<String> repoUrls, ProgressIndicator indicator) {
```
felixdesouza: this is not a safe change, if you add a url to `repoUrls` you're not going to be able to see updated results for at most 10 mins as you've keyed on metadata.

it doesn't form part of your cache key

I think it's fine for now since it's all constant, but we're going to have to think about the caching strategy when you're reading urls from your xml file.

roughly speaking, perhaps you want to cache just extracting all the strings from file regardless of validity, and then post-process with the urls outside of this method. If that then becomes a hot path, we can think about caching again there, but it's a bit awkward. Thoughts @CRogers ?

felixdesouza: it could even be a 'cache invalidate on gradle reload' which is kinda aligned with what you want i.e. 
* you potentially get a new `repositories.xml` when you reload gradle
* you potentially get a new binary file when you reload gradle

I suppose the only thing it doesn't capture is running gradle outside of intellij and seeing results in , idk

FinlayRJW: I've reverted these changes so that we can merge then we can figure this out in [this PR](https://github.com/palantir/gradle-consistent-versions-idea-plugin/pull/31) that add the repo reading

felixdesouza:
👍

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
CRogers: Invalidating when reloading gradle makes sense to me... probably also worth seeing just how long this takes to read.

FinlayRJW: to read and parse the file on my system is 62 milliseconds

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -36,27 +37,30 @@ class GradleCacheExplorerTest {
...
 
-        assertThat(explorer.isValidResourceUrl("https://example.com/com/example/artifact/1.0/artifact-1.0.pom"))
+        assertThat(explorer.isValidResourceUrl(
+                        projectUrls, "https://example.com/com/example/artifact/1.0/artifact-1.0.pom"))
                 .as("because the URL is not from a known valid repository")
                 .isFalse();
 
         assertThat(explorer.isValidResourceUrl(
-                        "https://repo.maven.apache.org/maven2/com/example/artifact/1.0/artifact-1.0.txt"))
+                        projectUrls, "https://repo.maven.apache.org/maven2/com/example/artifact/1.0/artifact-1.0.txt"))
```
CRogers: Is the disparity between the apache mirror and repo1.maven.org still apparent? Is this going to cause issues when reading in the `repositories.xml`?

FinlayRJW: from my testing the repo url that ends up in the `gcv-maven-repositrories.xml` matches the repos in the bin files

PR #37, opened 2024-10-09
Highlight dependency-upgrader comments

FinlayRJW:
## Before this PR
comments containing the special sequence `dependency-upgrader:OFF/ON` where treated as any other comment

## After this PR
We now have an annotator to highlight these comments making them easily identifiable 

image

==COMMIT_MSG==
Highlight dependency-upgrader comments
==COMMIT_MSG==


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/CommentAnnotator.java
```diff
@@ -0,0 +1,71 @@
...
+public class CommentAnnotator implements Annotator {
+
+    private static final TextAttributesKey RED_BOLD = TextAttributesKey.createTextAttributesKey(
+            "DEPENDENCY_UPGRADER_OFF", new TextAttributes(JBColor.RED, null, null, null, Font.BOLD));
+
+    private static final TextAttributesKey GREEN_BOLD = TextAttributesKey.createTextAttributesKey(
+            "DEPENDENCY_UPGRADER_ON", new TextAttributes(JBColor.GREEN, null, null, null, Font.BOLD));
+
+    private static final Pattern DEPENDENCY_UPGRADER_OFF_PATTERN =
+            Pattern.compile("dependency-upgrader:OFF", Pattern.CASE_INSENSITIVE);
```
felixdesouza: let's be a bit more lenient and allow any number of spaces between the colon and `OFF|ON`

internally there are some repos that have:

```props
# dependency-upgrader: OFF
```

FinlayRJW: fixed!

felixdesouza (approved):

PR #45, opened 2024-10-11
Cache is loaded on first instance and on project refresh

FinlayRJW:
## Before this PR
We loading the cache at completion time so closing a completion could create an empty cache that would not be reloaded till the cache was invalidated

## After this PR
Now the cache is an application service that is only loaded when the gradle project is refreshed to prevent it ever getting into a bad state 
==COMMIT_MSG==
Cache is loaded on first instance and on project refresh
==COMMIT_MSG==

## Possible downsides?
This has only been tested on my laptop with a fairly small gradle cache so may have performance issues on others laptops

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
@@ -63,7 +63,7 @@ public class GradleCacheExplorer {
         }
 
         stopWatch.stop();
-        log.warn("Cache parsing time: {} ms", stopWatch.getTime());
+        log.debug("Cache parsing time: {} ms", stopWatch.getTime());
 
         return results.stream()
                 .filter(result -> result.startsWith(parsedInput))
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
@@ -16,12 +16,7 @@
 
 package com.palantir.gradle.versions.intellij;
 
-import com.github.benmanes.caffeine.cache.Cache;
-import com.github.benmanes.caffeine.cache.Caffeine;
 import com.intellij.openapi.application.ApplicationManager;
-import com.intellij.openapi.progress.ProcessCanceledException;
-import com.intellij.openapi.progress.ProgressIndicator;
-import com.intellij.openapi.progress.ProgressManager;
 import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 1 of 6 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -31,90 +26,71 @@
...
 
     private static final Logger log = LoggerFactory.getLogger(GradleCacheExplorer.class);
     private static final String GRADLE_CACHE_PATH = System.getProperty("user.home") + "/.gradle/caches/modules-2/";
-    private final Cache<String, Set<String>> cache =
-            Caffeine.newBuilder().expireAfterWrite(10, TimeUnit.MINUTES).build();
+    private final AtomicReference<Set<String>> cache = new AtomicReference<>(Collections.emptySet());
 
-    public final void invalidateCache() {
-        cache.invalidateAll();
+    GradleCacheExplorer() {
```
FinlayRJW, line 42: Not sure if this is the best way to initialise the cache but seemed alot simpler than having an entire `backgroundPostStartupActivity`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -31,90 +26,71 @@
...
+            return results;
         }
 
-        try {
-            Callable<Set<String>> task = () -> extractStrings(repoUrls, indicator);
-            Future<Set<String>> future = ApplicationManager.getApplication().executeOnPooledThread(task);
-            Set<String> results =
-                    com.intellij.openapi.application.ex.ApplicationUtil.runWithCheckCanceled(future::get, indicator);
+        stopWatch.stop();
+        log.debug("Cache parsing time: {} ms", stopWatch.getTime());
```
CRogers: Can we call this "completion matching time" or something? "Cache parsing time" sounds like what `extractStrings()` is doing.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -31,90 +26,71 @@
...
+        loadCache();
+    }
+
+    public final void loadCache() {
+        cache.set(extractStrings());
     }
 
     public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup input) {
+        StopWatch stopWatch = new StopWatch();
+        stopWatch.start();
```
CRogers: `Stopwatch.createStarted()`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
CRogers: I suspect you may not be meant to do work in service init though - I wonder if it would slow down the application startup?

CRogers: Actually, this will happen when the service is created right? Which is probably going to be the first time someone asks for completions or refreshes the project, whichever is first? If so, that seems fine.

CRogers:
:+1:

PR #48, opened 2024-10-11
filter for rc and snapshot in versions

FinlayRJW:
## Before this PR
We would always get the latest version from the xml, however this may be an rc or snapshot version

## After this PR
Now we try and get the release version and fall back onto the latest version if we need to - if the release version contains rc / snapshot will will find the most recent version that doesnt to set as the latest version
==COMMIT_MSG==
filter for rc and snapshot in versions
==COMMIT_MSG==

## Possible downsides?
Extra compute time when getting the versions

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.dataformat.xml.XmlMapper;
 import com.fasterxml.jackson.datatype.guava.GuavaModule;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.annotations.VisibleForTesting;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
@@ -43,7 +44,7 @@ import org.slf4j.LoggerFactory;
 public class RepositoryExplorer {
     private static final Logger log = LoggerFactory.getLogger(RepositoryExplorer.class);
 
-    private static final Pattern VERSION_PATTERN =
+    private static final Pattern UNSTABLE_VERSION_PATTERN =
             Pattern.compile(".*(-rc\\d*|-SNAPSHOT|-M\\d+|-alpha|-beta)$", Pattern.CASE_INSENSITIVE);
 
     private final Cache<CacheKey, Set<GroupPartOrPackageName>> folderCache = Caffeine.newBuilder()
@@ -117,34 +118,41 @@ public class RepositoryExplorer {
             xmlMapper.registerModule(new GuavaModule());
 
             Metadata metadata = xmlMapper.readValue(content, Metadata.class);
-            List<String> allVersions = new ArrayList<>(metadata.versioning().versions());
-            Collections.reverse(allVersions);
-
-            if (allVersions.isEmpty()) {
-                return Collections.emptySet();
-            }
-
-            String latest = Optional.ofNullable(metadata.versioning().release())
-                    .filter(l -> !l.isEmpty())
-                    .orElseGet(() -> metadata.versioning().latest());
-
-            if (VERSION_PATTERN.matcher(latest.toLowerCase(Locale.ROOT)).matches()) {
-                latest = allVersions.stream()
-                        .filter(version -> !VERSION_PATTERN
-                                .matcher(version.toLowerCase(Locale.ROOT))
-                                .matches())
-                        .findFirst()
-                        .orElse(latest);
-            }
-
-            String finalLatest = latest;
-            return allVersions.stream()
-                    .map(version -> DependencyVersion.of(version, finalLatest.equals(version)))
-                    .collect(Collectors.toSet());
+            return parseVersionsFromContent(metadata);
         } catch (Exception e) {
             log.error("Failed to parse maven-metadata.xml", e);
         }
         return Collections.emptySet();
+    }
+
+    @VisibleForTesting
+    final Set<DependencyVersion> parseVersionsFromContent(Metadata metadata) {
+        List<String> allVersions = new ArrayList<>(metadata.versioning().versions());
+        Collections.reverse(allVersions);
+
+        if (allVersions.isEmpty()) {
+            return Collections.emptySet();
+        }
+
+        String metadataReleaseOrLatest = Optional.ofNullable(
+                        metadata.versioning().release())
+                .filter(l -> !l.isEmpty())
+                .orElseGet(() -> metadata.versioning().latest());
+
+        String latestStableVersion = Optional.of(metadataReleaseOrLatest)
+                .filter(this::isStableVersion)
+                .or(() -> allVersions.stream().filter(this::isStableVersion).findFirst())
+                .orElse(metadataReleaseOrLatest);
+
+        return allVersions.stream()
+                .map(version -> DependencyVersion.of(version, latestStableVersion.equals(version)))
+                .collect(Collectors.toSet());
+    }
+
+    private boolean isStableVersion(String version) {
+        return !UNSTABLE_VERSION_PATTERN
+                .matcher(version.toLowerCase(Locale.ROOT))
+                .matches();
     }
 
     @Value.Immutable
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
@@ -17,15 +17,20 @@
 package com.palantir.gradle.versions.intellij;
 
 import com.fasterxml.jackson.dataformat.xml.XmlMapper;
+import com.fasterxml.jackson.datatype.guava.GuavaModule;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import org.immutables.value.Value;
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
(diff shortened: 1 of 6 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -103,21 +108,39 @@ private Set<GroupPartOrPackageName> fetchFoldersFromContent(String contents) {
...
+
+            String latest = Optional.ofNullable(metadata.versioning().release())
+                    .filter(l -> !l.isEmpty())
+                    .orElseGet(() -> metadata.versioning().latest());
+
+            if (latest.toLowerCase(Locale.ROOT).contains("rc")
+                    || latest.toLowerCase(Locale.ROOT).contains("snapshot")) {
+                latest = allVersions.stream()
+                        .filter(version -> !version.toLowerCase(Locale.ROOT).contains("rc")
+                                && !version.toLowerCase(Locale.ROOT).contains("snapshot"))
```
CRogers: I'd check they end with `-rc*` or `-SNAPSHOT` (I could totally see an artifact having `rc` or `snapshot` in the name). Could probably use regex/Pattern. I think we'd want `-M1`, `-M2` as well as `-alpha` and `-beta`

CRogers: Oh wait, this is just checking the version. Probably still best to just check they end with.

FinlayRJW: fixed - now using regex match and looking for a wider range of matches

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -103,21 +112,39 @@ private Set<GroupPartOrPackageName> fetchFoldersFromContent(String contents) {
...
-                }
+            List<String> allVersions = new ArrayList<>(metadata.versioning().versions());
+            Collections.reverse(allVersions);
+
+            if (allVersions.isEmpty()) {
+                return Collections.emptySet();
             }
+
+            String latest = Optional.ofNullable(metadata.versioning().release())
+                    .filter(l -> !l.isEmpty())
```
felixdesouza: more descriptive variable name pls

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -103,21 +112,39 @@ private Set<GroupPartOrPackageName> fetchFoldersFromContent(String contents) {
...
+                    .orElseGet(() -> metadata.versioning().latest());
+
+            if (VERSION_PATTERN.matcher(latest.toLowerCase(Locale.ROOT)).matches()) {
```
felixdesouza: let's extract `VERSION_PATTERN.matcher(...).matches()` to a method:

```java
boolean isVersion(String maybeVersion) {
  return VERSION_PATTERN.matcher(latest.toLowerCase(Locale.ROOT)).matches();
}
```

then replace both usages of it

felixdesouza: I also don't understand the predicate:

> if `latest` is a well formed version, we find the first non-well formed version and use that as the latest?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -103,21 +112,39 @@ private Set<GroupPartOrPackageName> fetchFoldersFromContent(String contents) {
...
             XmlMapper xmlMapper = new XmlMapper();
+            xmlMapper.registerModule(new GuavaModule());
 
             Metadata metadata = xmlMapper.readValue(content, Metadata.class);
-            if (metadata.versioning() != null && metadata.versioning().versions() != null) {
-                String latest = metadata.versioning().latest();
-                for (String version : metadata.versioning().versions()) {
-                    versions.add(DependencyVersion.of(version, latest.equals(version)));
-                }
+            List<String> allVersions = new ArrayList<>(metadata.versioning().versions());
```
felixdesouza: this would do well with some tests, it's a bit hard to visualise what you're trying to show here.

To avoid having to serialise and deserialise xml, you make a helper method:

```java
@VisibleForTesting
Set<DependencyVersion> parseVersionsFromContent(Metadata metadata) {
 // ...
}
```

Then you can just make simple `Metadata` objects in your tests and test that things work the way you want

felixdesouza: it also acts as a way of codifying the ux that you want and prevent regressions are you evolve the ux more etc.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
FinlayRJW: Ah this is probably badly worded from me, the matcher is checking for verisons that are not good e.g. SNAPSHOT of beta versions, so if we latest is a snapshot version work backwards till we find the first version that is not a snapshot/beta/etc

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
FinlayRJW: yeah this whole section was very unclear I think it should now be alot clearer

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
FinlayRJW: I've tried to tidy this up so hopefully it is a lot more self explanatory now

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
FinlayRJW: Have added some tests so should be covered now

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -37,6 +44,9 @@
 public class RepositoryExplorer {
     private static final Logger log = LoggerFactory.getLogger(RepositoryExplorer.class);
 
+    private static final Pattern UNSTABLE_VERSION_PATTERN =
+            Pattern.compile(".*(-rc\\d*|-SNAPSHOT|-M\\d+|-alpha|-beta)$", Pattern.CASE_INSENSITIVE);
```
CRogers: I think we should match these as well:

```
-alpha2
-alpha-2
(same with beta)
-rc-1
```

As these are commonly used too.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -103,21 +113,48 @@ private Set<GroupPartOrPackageName> fetchFoldersFromContent(String contents) {
     }
 
     private Set<DependencyVersion> parseVersionsFromContent(String content) {
-        Set<DependencyVersion> versions = new HashSet<>();
         try {
             XmlMapper xmlMapper = new XmlMapper();
+            xmlMapper.registerModule(new GuavaModule());
```
CRogers: Can avoid GuavaModule here by setting `@Value.Style(jdkOnly = true)` on the immutables type. This is actually good practice; internally we have [a library](https://pl.ntr/2oV) for this that sets some good defaults.

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/RepositoryExplorerTests.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+public class RepositoryExplorerTests {
+
+    @Test
+    void test_gets_release_from_metadata() {
+        RepositoryExplorer repositoryExplorer = new RepositoryExplorer();
+
+        Versioning versioning = ImmutableVersioning.builder()
```
CRogers: Do not use `Immutable*` types outside the immutables classes. Make builds for them that returns the no `Immutable*` type (ie `static Versioning builder()`)

FinlayRJW: So I've set this up to work like in GCV where we return a builder, is this the correct thing todo? 

```java
    static Builder builder() {
        return new Builder();
    }

    class Builder extends ImmutableVersioning.Builder {}
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -38,6 +44,9 @@
 public class RepositoryExplorer {
     private static final Logger log = LoggerFactory.getLogger(RepositoryExplorer.class);
 
+    private static final Pattern UNSTABLE_VERSION_PATTERN = Pattern.compile(
+            ".*(-rc(\\d+|-\\d+)?|-SNAPSHOT|-M\\d+|-alpha(\\d+|-\\d+)?|-beta(\\d+|-\\d+)?)$", Pattern.CASE_INSENSITIVE);
```
CRogers: Every `\\d+|-\\d+` can be replaced by `-?\\d+`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -104,21 +113,49 @@ private Set<GroupPartOrPackageName> fetchFoldersFromContent(String contents) {
...
+                .orElseGet(() -> metadata.versioning().latest());
+
+        // Check if the releaseOrLatestVersion is stable, it not find first stable version, if no stable versions return
+        // the releaseOrLatestVersion
+        String latestStableVersion = Optional.of(releaseOrLatestVersion)
+                .filter(this::isStableVersion)
+                .or(() -> allVersions.stream().filter(this::isStableVersion).findFirst())
+                .orElse(releaseOrLatestVersion);
+
+        Collections.reverse(allVersions);
```
CRogers: Rather than reversing the `allVersions` list twice, can we just avoid reversing it mutably and use `Lists.reverse` where needed? This make a list "view" where the list is reversed without actually changing it.

CRogers (approved):

PR #51, opened 2024-10-14
Suggest the * wild card match for package names

FinlayRJW:
## Before this PR
The plugin never suggested using wild cards this is bad as it discourages the use of it

## After this PR
We now always will suggest the use of a * wildcard match for package name

We have also fixed a bug where groups contain a deeper group and packages e.g. you have typed `com.example.group:` and you get suggested `packageName1` and also `deeperGroup:otherPackageName`
 
==COMMIT_MSG==
Suggest the * wild card match for package names
==COMMIT_MSG==

## Possible downsides?
This means we now always suggest a * without checking if a * is possible for a particular project - we feel this is better than not suggesting a * at all though

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
@@ -23,6 +23,8 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Optional;
@@ -47,7 +49,7 @@ public class GradleCacheExplorer {
         cache.set(extractStrings());
     }
 
-    public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup input) {
+    public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup input, boolean isPackageName) {
         Stopwatch stopWatch = Stopwatch.createStarted();
 
         String parsedInput = String.join(".", input.parts());
@@ -57,17 +59,36 @@ public class GradleCacheExplorer {
                 .flatMap(Optional::stream)
                 .collect(Collectors.toSet());
 
+        Set<String> resultsWithStarsIncluded = results.stream()
+                .map(result -> {
+                    int colonIndex = result.indexOf(':');
+                    if (colonIndex != -1) {
+                        return Arrays.asList(result, result.substring(0, colonIndex) + ":*");
+                    }
+                    return Collections.singletonList(result);
+                })
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet());
+
         if (parsedInput.isEmpty()) {
             return results;
+        }
+
+        Set<String> filteredResults = resultsWithStarsIncluded.stream()
+                .filter(result -> result.startsWith(parsedInput))
+                .map(result -> result.substring(parsedInput.length() + 1))
+                .collect(Collectors.toSet());
+
+        if (isPackageName) {
+            filteredResults = filteredResults.stream()
+                    .filter(result -> !result.contains(":"))
+                    .collect(Collectors.toSet());
         }
 
         stopWatch.stop();
         log.debug("Completion matching time: {} ms", stopWatch.elapsed().toMillis());
 
-        return results.stream()
-                .filter(result -> result.startsWith(parsedInput))
-                .map(result -> result.substring(parsedInput.length() + 1))
-                .collect(Collectors.toSet());
+        return filteredResults;
     }
 
     private Set<String> extractStrings() {
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
@@ -16,6 +16,7 @@
 
 package com.palantir.gradle.versions.intellij;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
 import com.intellij.openapi.application.ApplicationManager;
 import java.io.BufferedInputStream;
@@ -42,6 +43,11 @@ public class GradleCacheExplorer {
 
     GradleCacheExplorer() {
         loadCache();
+    }
+
+    @VisibleForTesting
+    GradleCacheExplorer(Set<String> newCache) {
+        cache.set(newCache);
     }
 
     public final void loadCache() {
@@ -185,10 +191,5 @@ public class GradleCacheExplorer {
 
     static GradleCacheExplorer getInstance() {
         return ApplicationManager.getApplication().getService(GradleCacheExplorer.class);
-    }
-
-    // This should only ever be used by tests
-    final void setCacheForTesting(Set<String> newCache) {
-        cache.set(newCache);
     }
 }
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
@@ -64,9 +64,8 @@ public class GradleCacheExplorer {
                 .flatMap(Optional::stream)
                 .collect(Collectors.toSet());
 
-        Set<String> resultsWithStarsIncluded = results.stream()
-                .flatMap(result -> includeStarsIfMultipleArtifacts(result, results))
-                .collect(Collectors.toSet());
+        Set<String> resultsWithStarsIncluded =
+                results.stream().flatMap(GradleCacheExplorer::includeStars).collect(Collectors.toSet());
 
         if (parsedInput.isEmpty()) {
             return resultsWithStarsIncluded;
@@ -89,15 +88,11 @@ public class GradleCacheExplorer {
         return filteredResults;
     }
 
-    static Stream<String> includeStarsIfMultipleArtifacts(String result, Set<String> results) {
+    static Stream<String> includeStars(String result) {
         int colonIndex = result.indexOf(':');
         if (colonIndex != -1) {
             String prefix = result.substring(0, colonIndex);
-            long count =
-                    results.stream().filter(r -> r.startsWith(prefix + ":")).count();
-            if (count > 1) {
-                return Stream.of(result, prefix + ":*");
-            }
+            return Stream.of(result, prefix + ":*");
         }
         return Stream.of(result);
     }
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
@@ -23,8 +23,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Optional;
@@ -32,6 +30,7 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import one.util.streamex.StreamEx;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -57,17 +59,36 @@ public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup in
...
 
+        Set<String> resultsWithStarsIncluded = results.stream()
+                .map(result -> {
+                    int colonIndex = result.indexOf(':');
+                    if (colonIndex != -1) {
+                        return Arrays.asList(result, result.substring(0, colonIndex) + ":*");
+                    }
+                    return Collections.singletonList(result);
+                })
+                .flatMap(Collection::stream)
```
felixdesouza: you can replace the collection construction/instantiation with just a `Stream` instead and do a `flatMap` instead i.e.:

```java
.flatMap(result -> {
  int colonIndex = ...
  if (...) {
    return Stream.of(result, result.substring(..));
  }
  return Stream.of(result);
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -57,17 +59,36 @@ public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup in
...
+                .collect(Collectors.toSet());
+
         if (parsedInput.isEmpty()) {
             return results;
         }
```
felixdesouza: if this doesn't depend on `resultsWithStarsIncluded` you should move it above it

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -57,17 +59,36 @@ public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup in
...
+        Set<String> filteredResults = resultsWithStarsIncluded.stream()
                 .filter(result -> result.startsWith(parsedInput))
                 .map(result -> result.substring(parsedInput.length() + 1))
                 .collect(Collectors.toSet());
+
+        if (isPackageName) {
+            filteredResults = filteredResults.stream()
+                    .filter(result -> !result.contains(":"))
+                    .collect(Collectors.toSet());
+        }
```
felixdesouza: not the worst thing in the world, but a little strange that you're constructing and then deconstructing a set even though `isPackageName` is effectively static for the lifetime of this function.

You could add another filter to the stream above, but that seems "wasteful" in both cases i.e. if you `isPackageName` is false, you're still having that check before you get to your filter. Despite it always being `true` or `false` for the lifetime of this function.

A thing you could do is have a method: 

```java
Stream<...> packageNameFilter(Stream<...> stream, boolean isPackageName) {
  if (!isPackageName) {
    return stream;
  } else {
    return stream.filter(...);
  }
}

// ...
Set<String> filteredResults = ...;
return packageNameFilter(filteredResults, isPackageName);
```

If you use something like [StreamEx](https://github.com/amaembo/streamex) you can use their useful `chain` utility method which makes it more fluent:

```java
Set<String> filteredResults = StreamEx.of(resultsWithStarsIncluded)
  .filter(...)
  .map(...)
  .chain(packageNameFilter(...))
  .toSet();
```

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -57,17 +59,36 @@ public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup in
...
+
+        stopWatch.stop();
+        log.debug("Completion matching time: {} ms", stopWatch.elapsed().toMillis());
```
felixdesouza, line 86: I often prefer to return where possible over variable mutation as it's a bit easier to see where things end and what will be there at the end. This is harder now that you have the stopwatch and also the two different exit paths.

You can keep it in the class, but a method that will probably be useful is some sort of `timeOperation` method e.g.

```java
<T> T timeOperation(String operation, Supplier<T> supplier) {
  Stopwatch stopWatch = Stopwatch.createStarted();
  T result = supplier.get();
  stopWatch.stop();
  log.debug("Operation '{}' took {} ms", ...);
  return result; 
}
```

Then inside, you can return whenever, however you want. It matters a tiny bit less if you do the above suggestion since you'll only have one exit point.

If things were particularly performance sensitive, you would then do something like checking whether the `logger` had `debug` mode enabled, and if not, then avoid creating all the stopwatch/logging methods. This works nicely because you wouldn't have to change your calling code. It's likely not necessary for now, but something to be aware of.

felixdesouza (commented):
It would be good to have a test case for the thing you fixed + the new wildcard suggestion

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: good idea, I've also added a check to ensure that the count of packages is more than one as we dont want to suggest * for package name with only one option

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: Ah it should depend on `resultsWithStarsIncluded`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: switched to using streamx

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: good to know thanks!

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -182,4 +186,9 @@ Optional<String> extractGroupAndArtifactFromUrl(Set<String> repoUrls, String url
     static GradleCacheExplorer getInstance() {
         return ApplicationManager.getApplication().getService(GradleCacheExplorer.class);
     }
+
+    // This should only ever be used by tests
```
felixdesouza: there is a typically a `@VisibleForTesting` annotation for this

felixdesouza: prefer to use a constructor for this i.e.

```
GradleCacheExplorer(Set<String> cachedUrls) {
  cache.set(cachedUrls);
}
```

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -103,4 +103,92 @@ void test_extract_group_artifact_from_url_correctly() {
...
+                "https://example.one/com/exampleOne/artifactTwo/1.0/artifact-1.0.jar",
+                "https://example.one/com/exampleDifferent/artifactOne/1.0/artifact-1.0.jar",
+                "https://example.two/com/exampleTwo/artifactOne/1.0/artifact-1.0.pom",
+                "https://example.two/com/exampleTwo/artifactTwo/1.0/artifact-1.0.jar",
+                "https://example.one/org/exampleOrg/artifactOrg/1.0/artifact-1.0.pom",
+                "https://example.three/com/exampleThree/artifactOne/1.0/artifact-1.0.pom");
+
+        explorer.setCacheForTesting(cache);
+        assertThat(explorer.getCompletions(repoUrls, DependencyGroup.fromString(""), false))
+                .as("Matches all")
```
felixdesouza, line 129: matches all except `example.three` repository

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -103,4 +103,92 @@ void test_extract_group_artifact_from_url_correctly() {
...
+    void test_get_completion_match_some() {
+        Set<String> repoUrls = Set.of("https://example.one/", "https://example.two/");
+        Set<String> cache = Set.of(
+                "https://example.one/com/exampleOne/artifactOne/1.0/artifact-1.0.pom",
+                "https://example.one/com/exampleOne/artifactTwo/1.0/artifact-1.0.jar",
+                "https://example.one/com/exampleDifferent/artifactOne/1.0/artifact-1.0.jar",
+                "https://example.two/com/exampleTwo/artifactOne/1.0/artifact-1.0.pom",
+                "https://example.two/com/exampleTwo/artifactTwo/1.0/artifact-1.0.jar",
+                "https://example.one/org/exampleOrg/artifactOrg/1.0/artifact-1.0.pom",
+                "https://example.three/com/exampleThree/artifactOne/1.0/artifact-1.0.pom");
```
felixdesouza: number of urls makes it hard to see what you're trying to test and what things shouldn't be matched, let's:
* remove `example.three`, we have it covered by `test_completion_match_all`
* not fully representative, but just have `jar` or `pom` or have a helper method that spits out the two together and then flatten them

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -103,4 +103,92 @@ void test_extract_group_artifact_from_url_correctly() {
...
+        assertThat(explorer.getCompletions(repoUrls, DependencyGroup.fromString("com"), false))
+                .as("Matches all")
+                .containsOnly(
+                        "exampleOne:*",
+                        "exampleOne:artifactOne",
+                        "exampleOne:artifactTwo",
+                        "exampleDifferent:artifactOne",
+                        "exampleTwo:*",
+                        "exampleTwo:artifactOne",
+                        "exampleTwo:artifactTwo");
```
felixdesouza: it's a redundant assertion, but it's useful to be able to say:

```java
.doesNotContain("exampleOrg")
```

because all the strings are so similar, it's easy to get lost here

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
FinlayRJW: How will this work in the case where we are already using the constructor to load the cache - can I have multiple constructors? I essentially want a way to directly add to this classes cache as I dont think the runners have a real gradle cache I could use for tests

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
```diff
@@ -103,4 +103,92 @@ void test_extract_group_artifact_from_url_correctly() {
...
+        Set<String> repoUrls = Set.of("https://example.one/", "https://example.two/");
+        Set<String> cache = Set.of(
+                "https://example.one/com/exampleOne/artifactOne/1.0/artifact-1.0.pom",
+                "https://example.one/com/exampleOne/artifactTwo/1.0/artifact-1.0.jar",
+                "https://example.one/com/exampleOne/anotherGroup/artifactOne/1.0/artifact-1.0.jar",
+                "https://example.one/com/exampleDifferent/artifactOne/1.0/artifact-1.0.jar",
+                "https://example.two/com/exampleTwo/artifactOne/1.0/artifact-1.0.pom",
+                "https://example.two/com/exampleTwo/artifactTwo/1.0/artifact-1.0.jar",
+                "https://example.one/org/exampleOrg/artifactOrg/1.0/artifact-1.0.pom",
+                "https://example.three/com/exampleThree/artifactOne/1.0/artifact-1.0.pom");
```
felixdesouza: similarly here, let's trim down the number of things in the cache to make it easier to see what things you're asserting over

Also, to rule out any weird possibilties, name the artifacts in `exampleTwo` differently i.e. the things you have selected are "correct" because they're coming from `exampleOne` but how do you know that `exampleTwo` isn't polluting it?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -60,30 +59,35 @@ public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup in
...
                     int colonIndex = result.indexOf(':');
                     if (colonIndex != -1) {
-                        return Arrays.asList(result, result.substring(0, colonIndex) + ":*");
+                        String prefix = result.substring(0, colonIndex);
+                        long count = results.stream()
+                                .filter(r -> r.startsWith(prefix + ":"))
+                                .count();
+                        if (count > 1) {
+                            return Stream.of(result, prefix + ":*");
+                        }
```
felixdesouza: this seems new? what are you trying to show here?

FinlayRJW: I dont want to add a * suggestion if there is only one artifact to a group

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
felixdesouza: you can have multiple constructors

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
felixdesouza: let's extract the contents to a method and please add a test

FinlayRJW: I have extracted this and added tests, I have also simplified the previous tests to actually only test one thing at a time

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/GradleCacheExplorerTest.java
FinlayRJW: I have simplified the tests alot to only test one thing at a time so dont think this is needed anymore

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GradleCacheExplorer.java
```diff
@@ -57,17 +64,42 @@ public final Set<String> getCompletions(Set<String> repoUrls, DependencyGroup in
                 .flatMap(Optional::stream)
                 .collect(Collectors.toSet());
 
+        Set<String> resultsWithStarsIncluded = results.stream()
+                .flatMap(result -> includeStarsIfMultipleArtifacts(result, results))
```
CRogers: Honestly, I don't think we even need this logic for multiple artifacts. Even if there is just one artifact we should probably suggest `*`. At least for internal deps, I would prefer for people to start off with the `:*` as it means the star will be there when another package is added to that group.

FinlayRJW: I've made it always suggest star, and updated the tests to match this new behaviour

CRogers (approved):

PR #52, opened 2024-10-14
Maintain version order from xml

FinlayRJW:
## Before this PR
The order that versions appeared in as part of a suggestion would be determined by the `sort` order of the strings of the versions

## After this PR
We now use the `PrioritizedLookupElement` to force the order to be in that of the xml e.g. descending order
==COMMIT_MSG==
Maintain version order from xml
==COMMIT_MSG==

## Possible downsides?
If versions are added from a different contributor they will not automatically gain this ordering as it is added at the contributor level

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
@@ -21,10 +21,9 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.List;
+import java.util.LinkedHashSet;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
@@ -66,14 +65,14 @@ public class RepositoryExplorer {
         return fetchFoldersFromContent(content.get());
     }
 
-    public final List<DependencyVersion> getVersions(
+    public final Set<DependencyVersion> getVersions(
             DependencyGroup group, DependencyName dependencyPackage, String url) {
         String urlString = url + group.asUrlString() + dependencyPackage.name() + "/maven-metadata.xml";
         Optional<String> content = fetchContent(urlString);
 
         if (content.isEmpty()) {
             log.debug("Empty metadata content received");
-            return Collections.emptyList();
+            return Collections.emptySet();
         }
 
         return parseVersionsFromContent(content.get());
@@ -104,8 +103,8 @@ public class RepositoryExplorer {
         return folders;
     }
 
-    private List<DependencyVersion> parseVersionsFromContent(String content) {
-        List<DependencyVersion> versions = new ArrayList<>();
+    private Set<DependencyVersion> parseVersionsFromContent(String content) {
+        Set<DependencyVersion> versions = new LinkedHashSet<>();
         try {
             XmlMapper xmlMapper = new XmlMapper();
 
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
@@ -21,8 +21,10 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -102,8 +104,8 @@ private Set<GroupPartOrPackageName> fetchFoldersFromContent(String contents) {
         return folders;
     }
 
-    private Set<DependencyVersion> parseVersionsFromContent(String content) {
-        Set<DependencyVersion> versions = new HashSet<>();
+    private List<DependencyVersion> parseVersionsFromContent(String content) {
```
felixdesouza: I think you should keep this as a set, because now you're no longer checking for duplicates. Where you're using it below, is just converting it to some sort of iterable anyway, so this change doesn't do anything for you there.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -67,7 +70,14 @@ public void addCompletions(
                                                         .withLookupString("latest"),
                                                 Double.MAX_VALUE)
                                         : LookupElementBuilder.create(version))
-                                .forEach(resultSet::addElement);
+                                .toList();
+
+                        List<LookupElement> prioritizedVersions = IntStream.range(0, versions.size())
+                                .mapToObj(i ->
+                                        PrioritizedLookupElement.withPriority(versions.get(i), i - versions.size()))
+                                .toList();
```
felixdesouza: will this not override the priority set for the `latest` elements?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
felixdesouza: if you're looking for order preserving set, you could use `LinkedHashSet` instead

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
felixdesouza: if you used `StreamEx` from line 64, you can actually just zip with an infinite stream that counts down. 

Doesn't really matter if it starts at 0 since you want relative order

e.g. 
```java
StreamEx(...)
  .flatMap(...)
  .zipWith(IntStream.iterate(0, i -> i - 1)
  .map((version, priority) -> getLookupElement(version, priority)
  .forEach(resultSet::addElement);
```

felixdesouza (commented):
you should add some tests to verify the behaviour!

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
FinlayRJW: switched to streamex

FinlayRJW:
> you should add some tests to verify the behaviour!

I have added a test to confirm that the versions order is correct 😄

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -58,17 +61,22 @@ public void addCompletions(
...
                     }
+
+                    private LookupElement getLookupElement(DependencyVersion version, Integer priority) {
+                        return version.isLatest()
+                                ? PrioritizedLookupElement.withPriority(
+                                        LookupElementBuilder.create(version)
+                                                .withTypeText("Latest", true)
+                                                .withLookupString("latest"),
+                                        Double.MAX_VALUE)
+                                : PrioritizedLookupElement.withPriority(LookupElementBuilder.create(version), priority);
```
CRogers, line 78: This is making it so that if the repos are:

* jar
* dist


then all jar completions will appear before dist ones?

If so I'm not sure this is super necessary. It could be from either repo.

CRogers: Ah, it forces the order to be descending too, makes sense.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -58,17 +61,22 @@ public void addCompletions(
 
                         Project project = parameters.getOriginalFile().getProject();
 
-                        RepositoryLoader.loadRepositories(project).stream()
+                        StreamEx.of(RepositoryLoader.loadRepositories(project))
                                 .flatMap(url -> repositoryExplorer.getVersions(group, dependencyPackage, url).stream())
```
CRogers, line 65: Might be worth parallelising this somehow in another PR.

CRogers (approved):

Issue #57, opened 2024-10-17
No version completions if I put a space after the `=` in `versions.props`


CRogers:
## What happened?
When adding a pin in `versions.props`, if I add a space after the `=` I do not get completions. However, if I don't add the space, I get completions:

https://github.com/user-attachments/assets/58988a12-952d-4d82-81db-2d170513b1fe

## What did you want to happen?
To get the completions after doing a space.

FinlayRJW:
This should be solved in https://github.com/palantir/gradle-consistent-versions-idea-plugin/pull/61

PR #61, opened 2024-10-23
Naive star version suggestions

FinlayRJW:
## Before this PR
If a package had a star then in it we simply did not give version suggestions

## After this PR
For package names with star we now add all the versions for all the packages to the result set, we also use a custom weigher to ensure newer versions float to the top

## Possible downsides?
We are using internal intellij classes to allow us to refresh the results set - this is potentially dangerous although the function we are using has been the same for 12 years

We add all versions for all relevant packages, this can cause odd behaviour like 2 latest versions (from different packages)

We also have added a short lived version cache to make it more useable which could cause issues with not getting the true latest version

We dont handle * in the group


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -48,44 +48,59 @@ public class RepositoryExplorer {
     private static final Pattern UNSTABLE_VERSION_PATTERN = Pattern.compile(
             ".*(-rc(-?\\d+)?|-SNAPSHOT|-M\\d+|-alpha(-?\\d+)?|-beta(-?\\d+)?)$", Pattern.CASE_INSENSITIVE);
 
-    private final Cache<CacheKey, Set<GroupPartOrPackageName>> folderCache = Caffeine.newBuilder()
+    private final Cache<String, Set<GroupPartOrPackageName>> folderCache = Caffeine.newBuilder()
```
FinlayRJW: I realised that I can just use the full URL as the cache key rather than having a CacheKey object

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -48,44 +48,59 @@ public class RepositoryExplorer {
...
+    // can be very expensive to repeatedly get data that realistically doesn't change on a second by second basis so
+    // having a short-lived cache is okay
+    private final Cache<String, Set<DependencyVersion>> shortLivedVersionCache = Caffeine.newBuilder()
+            .expireAfterWrite(2, TimeUnit.MINUTES)
+            .maximumSize(10000)
+            .build();
+
     public final Set<GroupPartOrPackageName> getGroupPartOrPackageName(DependencyGroup group, String url) {
-        CacheKey cacheKey = CacheKey.of(url, group);
-        Set<GroupPartOrPackageName> folders = folderCache.get(cacheKey, key -> {
```
FinlayRJW, line 58: This could result in returning null if load folders was cancelled half way through

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -72,7 +98,7 @@ private LookupElement getLookupElement(DependencyVersion version, Integer priori
                         return version.isLatest()
                                 ? PrioritizedLookupElement.withPriority(
                                         LookupElementBuilder.create(version)
-                                                .withTypeText("Latest", true)
+                                                .withTypeText("Possible Latest", true)
```
FinlayRJW: Its always only possible as can have multiple repos with there own latest

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -86,6 +112,6 @@ private VersionPropsProperty findParentProperty(VersionPropsDependencyVersion ve
 
     @Override
     public final boolean invokeAutoPopup(PsiElement position, char typeChar) {
-        return typeChar == '=';
+        return true;
```
FinlayRJW, line 170: solves https://github.com/palantir/gradle-consistent-versions-idea-plugin/issues/57

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -48,47 +53,77 @@ public class RepositoryExplorer {
...
 
-    private Set<GroupPartOrPackageName> loadFolders(DependencyGroup group, String url) {
-        String urlString = url + group.asUrlString();
-        Optional<String> content = fetchContent(urlString);
+        Optional<String> content;
+        try {
+            content = fetchContent(urlString);
+        } catch (HttpException e) {
+            log.debug("Failed to fetch group/part/package", e);
+            folderCache.put(urlString, Collections.emptySet());
```
FinlayRJW, line 82: In this case we can cache the empty set as we know its from an empty bad url rather than from a cancel

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -158,14 +193,21 @@ private boolean isStableVersion(String version) {
                 .matches();
     }
 
-    @Value.Immutable
-    interface CacheKey {
-        String url();
+    private void triggerRefresh() {
```
FinlayRJW, line 196: We want to trigger the refresh at this level then we can refresh each time something is added to the cache

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsCodeInsightTest.java
```diff
@@ -52,31 +53,57 @@ public void test_version_completion() {
     }
 
     @Test
-    public void test_group_completion() throws Exception {
+    public void test_wild_card_version_completion() throws InterruptedException {
+        JavaCodeInsightTestFixture fixture = getFixture();
+        // The file name is required for context but does not need to exist on the filesystem
+        fixture.configureByText("versions.props", "com.palantir.tokens:* = <caret>");
+        fixture.complete(CompletionType.BASIC);
+        Thread.sleep(500);
```
CRogers: Thread.sleep in tests like this is often risky and can make flaky tests. Generally the way to do this is to use [awaitility](https://github.com/awaitility/awaitility) to wait up to say 1 min but try every X milliseconds until it works. It keeps tests running quickly but also means if something takes a unexpectedly long time time (maybe a network request takes 5 seconds?) the test does not fail and so is (much less) flaky

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RefreshUtil.java
```diff
@@ -0,0 +1,59 @@
...
+import com.intellij.codeInsight.completion.BaseCompletionService;
+import com.intellij.codeInsight.completion.CompletionProcess;
+import com.intellij.codeInsight.completion.CompletionProgressIndicator;
+import com.intellij.codeInsight.completion.CompletionService;
+import com.intellij.openapi.application.ApplicationManager;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RefreshUtil {
```
CRogers: I'd probably call this `CompletionRefreshUtil` - refresh alone isn't super clear.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RefreshUtil.java
```diff
@@ -0,0 +1,59 @@
...
+            triggerRefresh();
+            return null;
+        });
+    }
+
+    private static void triggerRefresh() {
+        ApplicationManager.getApplication().invokeLater(() -> {
+            CompletionService completionService = CompletionService.getCompletionService();
+            if (completionService == null) {
+                return;
```
CRogers: I would throw something like an `IllegalStateException` here. Otherwise if this changes/breaks then the plugin just silently stops working. We want lots of errors so it shows up in logs etc and we can work out what's gone wrong (and people give us useful bug reports).

CRogers: Actually, for this one maybe it makes sense if this goes away when completion ends?? But definitely if we expect this to always exist.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RefreshUtil.java
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            BaseCompletionService baseCompletionService = (BaseCompletionService) completionService;
```
CRogers: Similarly, we should throw here if it's not the correct type. You can use some nifty java 17 stuff now:

```java
if (!(completionService instanceof BaseCompletionService baseCompletionService)) {
    throw new IllegalStateException(
            "Expected completionService to be an instance of BaseCompletionService");
}
```

which allows you to use `baseCompletionService` below.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/ContentsUtil.java
```diff
@@ -16,65 +16,63 @@
...
-
-        if (indicator == null) {
-            return Optional.empty();
+    public static ContentResults fetchPageContents(String urlString) {
+        try {
+            URL url = new URL(urlString);
+            return ContentsUtil.fetchPageContents(url);
+        } catch (MalformedURLException e) {
+            log.error("Malformed URL", e);
+            return ContentResults.empty();
```
CRogers: I'd throw here, if we're ever making invalid URLs something has gone wrong in our code

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/ContentsUtil.java
```diff
@@ -16,65 +16,63 @@
...
 
                 BufferedReader in =
                         new BufferedReader(new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
-
-                return in.lines().collect(Collectors.joining("\n"));
+                return ContentResults.of(in.lines().collect(Collectors.joining("\n")));
             } catch (ConnectException e) {
-                log.debug("Connection refused on page {}", pageUrl, e);
-                return null;
+                log.warn("Connection refused on page {}", pageUrl, e);
```
CRogers: I had only made this warn as to make it show up - should probably still be debug as it's an expected thing to happen.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/ContentsUtil.java
```diff
@@ -16,65 +16,63 @@
...
+                int responseCode = connection.getResponseCode();
+                if (responseCode != HttpURLConnection.HTTP_OK) {
+                    return ContentResults.error(responseCode);
                 }
 
                 BufferedReader in =
                         new BufferedReader(new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
-
-                return in.lines().collect(Collectors.joining("\n"));
+                return ContentResults.of(in.lines().collect(Collectors.joining("\n")));
```
CRogers: Getting the lines then joining them back together is kinda too much effort. I would do `CharStreams.toString(inputStreamReader)`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GroupPartOrPackageNameExplorer.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.nodes.Element;
+import org.jsoup.select.Elements;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GroupPartOrPackageNameExplorer {
```
CRogers: Definitely better split out.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -29,15 +30,22 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import one.util.streamex.StreamEx;
 
 public class VersionCompletionContributor extends CompletionContributor {
-
-    private static final RepositoryExplorer repositoryExplorer = new RepositoryExplorer();
+    private static final GroupPartOrPackageNameExplorer GROUP_PART_OR_PACKAGE_NAME_EXPLORER =
+            new GroupPartOrPackageNameExplorer();
```
CRogers: This means there's two `GroupPartOrPackageNameExplorer` instances, one in this completer, one in the other. Can we make these `Explorer` classes services somehow so there's only ever one instance of them?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -47,45 +55,114 @@ public class VersionCompletionContributor extends CompletionContributor {
...
+        RepositoryLoader.loadRepositories(project)
+                .forEach(url -> addToResults(sortedResultSet, List.of(new GroupAndDep(group, dependencyName, url))));
+    }
+
+    private List<GroupAndDep> collectGroupAndDeps(
+            Project project, DependencyGroup group, DependencyName dependencyName) {
+        String dependencyNamePrefix = dependencyName.name().replace("*", "");
+        return StreamEx.of(RepositoryLoader.loadRepositories(project))
+                .flatMap(url -> StreamEx.of(GROUP_PART_OR_PACKAGE_NAME_EXPLORER.getGroupPartOrPackageName(
+                                group, url, RefreshUtil.refreshOnceSupplier()::get))
```
CRogers: I don't think there's any point in scheduling a refresh here as we haven't loaded any versions into the completions yet. Similarly for the package completions, we probably don't need to refresh as we don't care so much about the ordering like we do with versions (versions we always want latest to be in the right place, packages people are typing partial parts of the name which is more "normal").

CRogers: I'd probably just make this block as it did before, maybe loading everything in parallel. Especially as most of the time it's just parsing a whole load of links on a single html page?

CRogers: Maybe it would make more sense for a `com.*:*` pattern when there are multiple things to load.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionExplorer.java
```diff
@@ -34,83 +32,62 @@
...
+            .maximumSize(10000)
+            .buildAsync(this::fetchAndParseFromUrl);
 
-    public final Set<GroupPartOrPackageName> getGroupPartOrPackageName(DependencyGroup group, String url) {
-        CacheKey cacheKey = CacheKey.of(url, group);
-        Set<GroupPartOrPackageName> folders = folderCache.get(cacheKey, key -> {
-            Set<GroupPartOrPackageName> loadedFolders = loadFolders(key.group(), url);
-            return loadedFolders.isEmpty() ? null : loadedFolders;
-        });
+    public record GroupAndDep(DependencyGroup group, DependencyName dependencyPackage, String url) {}
```
CRogers: This was just a random name I gave this when hacking it together. I'd probably call it something better - PackageInRepo or something like that?

`url` should be something like `repositoryUrl` - perhaps even with it's own type?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionExplorer.java
```diff
@@ -34,83 +32,62 @@
...
-        String urlString = url + group.asUrlString();
-        Optional<String> content = fetchContent(urlString);
+        Optional<Set<DependencyVersion>> versionsOptional =
+                Optional.ofNullable(shortLivedVersionCache.synchronous().getIfPresent(urlString));
 
-        if (content.isEmpty()) {
-            log.debug("Page does not exist");
-            return Collections.emptySet();
+        if (versionsOptional.isEmpty()) {
+            log.debug("Not loaded group: {}", groupAndDep);
```
CRogers: Can probably get rid of this.

gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsCodeInsightTest.java
FinlayRJW: good to know thanks!

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RefreshUtil.java
FinlayRJW: I'm pretty sure this should always exist, I've played around a bunch and we never seem to go into this state so probably best to throw

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
FinlayRJW: converted to application services similar to how the cache explorer works

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
FinlayRJW: So I have converted it to being synchronous again and added the progress indicator logic back in for the `GroupPartOrPackageNameExplorer` to prevent it blocking the UI - I think a future PR can worry about `*:*`

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionExplorer.java
FinlayRJW: renamed and created a repo url type

CRogers:
:+1:

FinlayRJW:
👍

Issue #47, opened 2024-10-11
Don't run writeVersionLocks when checking out a new branch


jackwickham:
## What happened?

When checking out a branch (using the terminal) which changes versions.props and versions.locks, the plugin triggers writeVersionLocks to run redundantly.

## What did you want to happen?

It should not have needed to run it. Ideally it would notice that versions.lock also got updated at the same time, so there's no need to rerun it.

PR #67, opened 2024-11-01
Don't write locks if both the locks and props change

FinlayRJW:
## Before this PR
Issue https://github.com/palantir/gradle-consistent-versions-idea-plugin/issues/47 outlined how changing branch triggered a write lock if both props and locks where different

## After this PR
We now track both props and locks changes and only trigger locks writes if the locks file has not changed within 350ms of the props file changing

This was slightly more complex than expected as the events that `prepareChange` are batched so you cannot guarantee that the change to props and locks will be in the same call to `prepareChange` so you must track state

==COMMIT_MSG==
Don't write locks if both the locks and props change
==COMMIT_MSG==

## Possible downsides?
This could prevent intentional locks writing if the props and locks change within 350ms of one another


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
```diff
@@ -47,6 +52,14 @@ public final class VersionPropsFileListener implements AsyncFileListener {
     private static final Logger log = LoggerFactory.getLogger(VersionPropsFileListener.class);
     private static final String TASK_NAME = "writeVersionsLock";
 
+    // Shared state to track changes per project
+    private final ConcurrentMap<Project, ChangeFlags> projectChangeMap =
+            new MapMaker().weakKeys().makeMap();
+
+    // Executor for debouncing
```
CRogers: I just did a bit of experimentation and it looks like the currently file you're editing (not just open) always gets it's file events in it's own batch. If you aren't actively editing versions.props/lock, the non-debounced version works fine, which would honestly probably solve most of the problem.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
```diff
@@ -47,6 +52,14 @@ public final class VersionPropsFileListener implements AsyncFileListener {
...
+    private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
```
CRogers: I think this implementation could be improved from a code design perspective. This combines debouncing with the rest of the logic for actually regenerating the locks and makes it do too much/is more confusing. Especially the state in `projectChangeMap`.

Instead, I would try to entirely separate out the debouncing behaviour from the other behaviour. The way I'd do this is keep the `VersionPropsFileListener` as is (with the don't run if `versions.lock` is in the same batch behaviour added). I would then make a `DebouncingAsyncFileListener` that takes a delegate `AsyncFileListener` in it's constructor. It builds up a batch of debounced events that it then finally gives to the delegate file listener once the debounce waiting period has expired. (You're probably also going to have make another `AsyncFileListener` to construct the whole combination of `new DebouncedAFL(new VPLF())` you can put in the `plugin.xml`).

This way the debouncing behaviour is fully distinct from the the rest of the code. There's less complicated state for `VersionPropsFileListener` like the `projectChangeMap` as you can just inspect the batch without it needing to worry about the debouncing behaviour and previous batches. You've also made a generally useful debouncing class that can be used elsewhere. The debouncing class can also be deleted very easily without having to change other classes in a significant way - this is ability to easily _delete_ functionality is normally a sign of good design.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
FinlayRJW: Interesting! In all my testing I had props open and editing so I was always getting it in its own batch - is it worth simplifying down then as it will only double run if the user has props/locks open and editing?

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileListener.java
FinlayRJW: I have now switched to using a separate debounce class and have used the `AppLifecycleListener` to register the listeners at application startup - I considered using a application service but this required me to use `preload` which is internal only

FinlayRJW: I have now switched to use `AsyncFileListener` to register the debounced listener as it is much simpler

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DebouncingAsyncFileListener.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class DebouncingAsyncFileListener implements AsyncFileListener, Disposable {
+    private static final Logger log = LoggerFactory.getLogger(DebouncingAsyncFileListener.class);
+
+    private final AsyncFileListener delegate;
+    private final SingleAlarm alarm;
```
FinlayRJW, line 36: Intellij class that ensures only a single instance of the task is scheduled at any one time

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsListenerRegistrar.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import org.jetbrains.annotations.Nullable;
+
+public final class VersionPropsListenerRegistrar implements AsyncFileListener {
+
+    private final DebouncingAsyncFileListener debouncedListener;
+
+    VersionPropsListenerRegistrar() {
+        AsyncFileListener versionPropsListener = new VersionPropsFileListener();
+        debouncedListener = new DebouncingAsyncFileListener(versionPropsListener, 250);
```
CRogers: I would probably write this all in one line, also a general style thing is to always use `this.` in constructors for referencing fields (you're normally forced to do this by the constructor params having the same name, but it looks odd if you don't for everything - in constructors only).

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DebouncingAsyncFileListener.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public void dispose() {
+        synchronized (lock) {
+            if (isDisposed) {
+                return;
+            }
+            isDisposed = true;
+            alarm.cancelAllRequests();
+        }
+    }
```
CRogers: just curious, was there some docs that recommend do this?

FinlayRJW: No docs but `SingleAlarm` needs a disposal parent and I was originally going to use the application as the disposal object but the docs say never todo that (https://plugins.jetbrains.com/docs/intellij/disposers.html#choosing-a-disposable-parent) so I had to implement my own disposal

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DebouncingAsyncFileListener.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class DebouncingAsyncFileListener implements AsyncFileListener, Disposable {
+    private static final Logger log = LoggerFactory.getLogger(DebouncingAsyncFileListener.class);
+
+    private final AsyncFileListener delegate;
```
CRogers, line 35: In general, if locking an object using the `synchronized` keyword it's best to [annotate the field with `@GuardedBy` from errorprone](https://errorprone.info/bugpattern/GuardedBy). This will ensure that any usages of the field are locked appropriately. It's ever so easy to modify the class and forget to do this (and suddenly everything breaks in the worse possible way).

But it's also generally better to use some type that intrinsically thread safe, which `ArrayList` is not. The best option here is likely a [`BlockingQueue`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/BlockingQueue.html) implementation (`BlockingQueue` is fantastic and generally the go to primitive when dealing with concurrent code). It has a [`drainTo` ](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/BlockingQueue.html#drainTo(java.util.Collection):~:text=int%C2%A0drainTo,E%3E%C2%A0c) method which allows you to just grab all of the elements from the queue in a thread safe way. Of the implementations, `ArrayBlockingQueue` is more performant, but it has a fixed capacity and we don't know how many elements there will be, so I'd use a `LinkedBlockingQueue`.

CRogers: Tbh, performance likely doesn't really matter here, but another improvement here might be to make a `FilteringAsyncFileListener` that just removes everything except the versions.props and versions.locks events before it gets to this stage. Might be nice just to build just to ensure we don't end up batching a huge number of events we don't care about. Can make it general and just have a `Predicate<VFileEvent>` that is configured in the ...Registrar class.

CRogers: (I'm assuming here `SingleAlarm` is threadsafe and doesn't need locking).

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DebouncingAsyncFileListener.java
CRogers: I'm not sure this is going to do anything. This class is now `Disposable` but what calls `dispose` on it? IntelliJ doesn't know about this class as it's encapsulated in `VersionPropsListenerRegistrar`. I'd expect for it to work properly `VersionPropsListenerRegistrar` would need to be `Disposable` and call dispose on `debouncedListener` when it itself is `dispose`d. If you really wanted to be safe you'd probably also check whether the `delegate` here is `Disposable` and then call `dispose` on it (or rely on `VersionPropsListenerRegistrar` disposing it). It all seems very manual/error prone.

FinlayRJW: I've fixed this up such that `VersionPropsListenerRegistrar` is disposable and this is passed as the parent down into the `SingleAlarm` class

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DebouncingAsyncFileListener.java
FinlayRJW: I've switched to using `LinkedBlockingQueue` and have added a general `FilteringAsyncFileListener` which is used to filter out all but props and locks changes

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsListenerRegistrar.java
```diff
@@ -16,22 +16,35 @@
...
-public final class VersionPropsListenerRegistrar implements AsyncFileListener {
+public final class VersionPropsListenerRegistrar implements AsyncFileListener, Disposable {
 
-    private final DebouncingAsyncFileListener debouncedListener;
+    private final FilteringAsyncFileListener changeListener;
 
     VersionPropsListenerRegistrar() {
-        this.debouncedListener = new DebouncingAsyncFileListener(new VersionPropsFileListener(), 250);
+
+        Predicate<VirtualFile> filter = virtualFile -> {
```
CRogers: I'd make this a method you can refer to by method reference

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DebouncingAsyncFileListener.java
```diff
@@ -23,63 +23,44 @@
...
 
-    private final List<VFileEvent> bufferedEvents = new ArrayList<>();
-    private boolean isDisposed = false;
-
-    DebouncingAsyncFileListener(AsyncFileListener delegate, int debounceDelayMillis) {
+    DebouncingAsyncFileListener(AsyncFileListener delegate, int debounceDelayMillis, Disposable parentDisposable) {
         this.delegate = delegate;
-        this.alarm = new SingleAlarm(this::processEvents, debounceDelayMillis, this, Alarm.ThreadToUse.POOLED_THREAD);
+        this.alarm = new SingleAlarm(
+                this::processEvents, debounceDelayMillis, parentDisposable, Alarm.ThreadToUse.POOLED_THREAD);
```
CRogers, line 42: This way of passing disposables is much better.

gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DebouncingAsyncFileListener.java
```diff
@@ -23,63 +23,44 @@
...
 import org.slf4j.LoggerFactory;
 
-public final class DebouncingAsyncFileListener implements AsyncFileListener, Disposable {
+public final class DebouncingAsyncFileListener implements AsyncFileListener {
     private static final Logger log = LoggerFactory.getLogger(DebouncingAsyncFileListener.class);
 
     private final AsyncFileListener delegate;
     private final SingleAlarm alarm;
-    private final Object lock = new Object();
+    private final LinkedBlockingQueue<VFileEvent> bufferedEvents = new LinkedBlockingQueue<>();
```
CRogers: Just use `BlockingQueue` instead of `LinkedBlockingQueue` when referring to it. The same as when we use `List` instead of `ArrayList`. This makes it easier to change the implementation later if we need to.

CRogers:
:+1:

PR #71, opened 2024-11-05
Order repositories based on url contents

FinlayRJW:
## Before this PR
The repos where in whatever order they happened to be in from `gcv-maven-repositories.xml` this was often not ideal as the most common repos where not first and all the others had to be checked first

## After this PR
Now we use a comparator to give preference to repo urls that contain `release` or `jar` and give less preference to repos with `internal` or `dist`

Tests where also added for the reader class
==COMMIT_MSG==
Order repositories based on url contents 
==COMMIT_MSG==

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/RepositoryLoaderTests.java
+++ b/gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/RepositoryLoaderTests.java
@@ -1,0 +1,110 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


gradle-consistent-versions-idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/RepositoryLoaderTests.java
```diff
@@ -0,0 +1,157 @@
...
+                        "jar",
+                        "internal-jar",
+                        "random",
+                        "dist",
+                        "internal",
+                        "internal-dist");
+    }
+
+    @Test
+    void repos_not_sorted_maintain_entry_order() throws IOException {
```
CRogers, line 129: nice

CRogers (approved):

PR #79, opened 2024-11-14
Fix plugin verifier issues

FinlayRJW:
## Before this PR
We were using an internal intellij API which was being detected by intellij (see comment in diff for why)

## After this PR
We are now calling this method in a way that prevents detection (see comment in diff)

==COMMIT_MSG==
Fix plugin verifier issues using
==COMMIT_MSG==

## Possible downsides?
We should try to find an alternative that doesn't use the internal API long term as this is a bit of a hack and the internal API could change


CRogers:
:+1:

FinlayRJW:
👍

PR #82, opened 2024-11-14
Parser correctly recovers on error

FinlayRJW:
## Before this PR
If a line had an error it would cause all other lines after it to break

## After this PR
We now use `recoverWhile` to recover after a parsing error fixing this issue

==COMMIT_MSG==
Parser correctly recovers on error
==COMMIT_MSG==


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.bnf
```diff
@@ -33,7 +33,10 @@ simpleFile ::= item_*
 
 private item_ ::= (property | COMMENT | CRLF)
 
-property ::= dependencyGroup COLON dependencyName EQUALS dependencyVersion
+property ::= dependencyGroup COLON dependencyName EQUALS dependencyVersion { recoverWhile="propertyRecover" }
```
FinlayRJW, line 36: This took me far to long to find!

CRogers: Very nice this exists!

CRogers (dismissed):

CRogers:
:+1:

PR #87, opened 2024-11-20
Allow for . character in package name during lexing

FinlayRJW:
## Before this PR
We where using the same regex for group parts and package names which did not include the . character, this break lexing if the package name includes a .

## After this PR
We now use separate regex's so that . can be included in the package name
==COMMIT_MSG==
Allow for . character in package name during lexing
==COMMIT_MSG==

### DIFF
--- a/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
+++ b/gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
@@ -42,17 +42,18 @@ VERSION=[^= \n\f#]+
 COLON=[:]
 EQUALS=[=]
 DOT=[.]
-IDENTIFIER = [^.:=\ \n\t\f]+
+GROUP_PART = [^.:=\ \n\t\f]+
+NAME_KEY = [^:=\ \n\t\f]+
 COMMENT=("#")[^\r\n]*
 
 %%
 
 <YYINITIAL> {WHITE_SPACE}*{COMMENT}              { return VersionPropsTypes.COMMENT; }
-<YYINITIAL> {IDENTIFIER}                         { return VersionPropsTypes.GROUP_PART; }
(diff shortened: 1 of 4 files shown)


gradle-consistent-versions-idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex
```diff
@@ -42,17 +42,18 @@ VERSION=[^= \n\f#]+
 COLON=[:]
 EQUALS=[=]
 DOT=[.]
-IDENTIFIER = [^.:=\ \n\t\f]+
+GROUP_PART = [^.:=\ \n\t\f]+
+NAME_KEY = [^:=\ \n\t\f]+
```
CRogers, line 46: Might as well add a lexer test for this too? At least we'll see a change if someone changes the lexer wrong in future.

I also verified it fixes the issue manually:

Screenshot 2024-11-20 at 13 56 15

FinlayRJW: Good idea added a test 😄

gradle-consistent-versions-idea-plugin/src/test/resources/lexerTests/93f5eb3eb699df3086a76de218626686409862745b5ac17aba383549f81708e4
```diff
@@ -0,0 +1,19 @@
...
+VersionPropsTokenType.GROUP_PART this
+VersionPropsTokenType.DOT .
+VersionPropsTokenType.GROUP_PART is
+VersionPropsTokenType.DOT .
+VersionPropsTokenType.GROUP_PART a
+VersionPropsTokenType.DOT .
+VersionPropsTokenType.GROUP_PART test
+DEPENDENCY_GROUP this.is.a.test
+VersionPropsTokenType.COLON :
+VersionPropsTokenType.NAME_KEY with.dots.in.the.name
```
CRogers, line 11: Nice

CRogers (approved):

